PR #116, opened 2022-09-19
PIN - 1851 

Carminepo2:
This bug was due to the fact that in case an action needed a confirm dialog, the *runAction* function did not "await" the user confirmation but it synchronously just returned void.

```javascript
if (showConfirmDialog) {
  wrapActionInDialog(runBasicAction, request.path.endpoint)
} else {
  return await runBasicAction()
}
// returns void
```
So when an action with the confirmation dialog is fired it instantly resolves, without waiting for the actual async action to resolve.

This led to a crash inside *wrapCreateNewVersionDraft* function (*AsyncTableEservice.tsx*) due to an object decostruction on a function that returns void.

```javascript
const { outcome, response } = (await runAction(
  ...
```

---
The proposed solution transforms *wrapActionInDialog* in an async function that resolves only when the user confirms or cancels the action.

### DIFF
--- a/src/hooks/useFeedback.tsx
+++ b/src/hooks/useFeedback.tsx
@@ -191,9 +191,9 @@ export const useFeedback = () => {
 
     if (showConfirmDialog) {
       return await wrapActionInDialog(runBasicAction, request.path.endpoint)
-    } else {
-      return await runBasicAction()
     }
+
+    return await runBasicAction()
   }
   /*
    * End API calls
--- a/src/hooks/useFeedback.tsx
+++ b/src/hooks/useFeedback.tsx
@@ -59,31 +59,25 @@ export const useFeedback = () => {
 
     if (!hasDialog) {
       throw new Error('This action should have a modal')
-    } else {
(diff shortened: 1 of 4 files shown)


src/hooks/useFeedback.tsx
```diff
@@ -179,7 +190,7 @@ export const useFeedback = () => {
     const runBasicAction = wrapBasicAction(request, options)
 
     if (showConfirmDialog) {
-      wrapActionInDialog(runBasicAction, request.path.endpoint)
+      return await wrapActionInDialog(runBasicAction, request.path.endpoint)
     } else {
       return await runBasicAction()
```
ruggerocastagnola, line 195: Now that it is symmetrical, can we also remove the `else`?

```
if(showConfirmDialog) {
  return await wrapActionInDialog(runBasicAction, request.path.endpoint)
}

return await runBasicAction()
```

ruggerocastagnola (changes requested):

ruggerocastagnola (approved):

PR #128, opened 2022-10-03
Pin 1932 - Redesign e-service details subscriber side

Carminepo2:
This branch was created from the [pin-1704 branch](https://github.com/pagopa/pdnd-interop-frontend/tree/pin-1704).

Major changes:

- Brought UI redesign E-Service details ([PIN-1704](https://github.com/pagopa/pdnd-interop-frontend/pull/127)) in the subscriber context. 
- Added InformationRow **content container automatic adaptation** based on the type of its children ("div" if the children contains ReactNodes, else "p").

### DIFF
--- a/src/views/EServiceManage.tsx
+++ b/src/views/EServiceManage.tsx
@@ -1,49 +1,28 @@
-import React, { FormEvent, useMemo, useState } from 'react'
+import React from 'react'
 import { StyledIntro } from '../components/Shared/StyledIntro'
 import {
   ActionProps,
   ApiEndpointKey,
   EServiceDescriptorRead,
-  EServiceDocumentRead,
   EServiceReadType,
   EServiceState,
-  FrontendAttributes,
 } from '../../types'
 import { RunActionOutput, useFeedback } from '../hooks/useFeedback'
 import { useHistory, useParams } from 'react-router-dom'
-import {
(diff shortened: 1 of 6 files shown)


src/views/EServiceManage.tsx
```diff
@@ -255,31 +231,39 @@ export function EServiceManage() {
         </Stack>
       </Stack>
 
-      {eserviceData && (
+      {!isLoading && eserviceData && descriptorId ? (
```
ruggerocastagnola, line 234: why `descriptorId` instead of `mode==="provider"`?

ruggerocastagnola (changes requested):

src/views/EServiceManage.tsx
Carminepo2: This line is part of the conditional render of the content/spinner based on the loading state. 

"**eserviceData && descriptorId**" is there just because I didn't want to put the non-null assertion operator in the props passed to the EServiceContentInfo, so I checked for them being non-null there.

It might not be clear, do you prefer to put the non-null assertion to the component's props?

```jsx
<EServiceContentInfo
  data={eserviceData!}
  descriptorId={descriptorId!}
  context="provider"
/>
```  

N.B: asserting that those variable are not null is correct in this context.

PR #212, opened 2022-12-19
[PIN-2478] Preparation testing router logic

Carminepo2:
This PR aims to make the routing code clearer and testable, changing files and functions naming, adding jsDocs and removing unused lines.

- AUTH_LEVEL route property is no longer optional;
- Removed unused property EXACT from route types;
- Added a runtime check that checks for localized paths consistency. It throws if the localized paths have different dynamic path names, or have different segments length;
- Added `react-i18next` testing function stubs;
- Added EServiceReadType mock factory function;
- Added utilities in `@/__mocks__/mock.utils.ts` to easily render a configurable application context in tests;
- Removed log "pagopa_env not available" on tests;
- Added check session expiration in AuthGuard;


src/router/router.types.ts
```diff
@@ -33,13 +33,13 @@ export type Navigate = <T extends RouteKey>(
...
-export type RouteConfig = {
+
+export type LocalizedRoute = {
   PATH: Record<LangCode, Readonly<string>>
   LABEL: Record<LangCode, Readonly<string>>
-  EXACT?: boolean
   COMPONENT: React.FC
   PUBLIC: boolean
-  AUTH_LEVELS?: RouteAuthLevel
+  AUTH_LEVELS: RouteAuthLevel
```
ruggerocastagnola, line 42: @Carminepo2 let's remove `any` as a possible value, always list all the `AUTH_LEVELS` allowed to see the content (`any === ['admin', 'api', 'security']`

ruggerocastagnola (approved):

PR #250, opened 2023-01-30
[PIN - 2405] - useActiveTab refactor and testing

alten-dturus:


Carminepo2 (changes requested):
Missing url params preservation on tab change and url params/state synchronization tests

ruggerocastagnola (approved):

PR #277, opened 2023-02-20
[PIN-1958] - Input validation and sanitization

alten-dturus:
- Removed Yup validation (and removed `yup` library and package `@hookform/resolvers`)
- Added parameters for react-hook-form validation in form components
- Added validation rules in forms
- Implemented a map function that map the react-hook-form validation error to different default error messages
- Implemented a deepTrim function that performs a trim operation on each string contained in an object and added that function in the axios request interceptor (for the request payload)


src/config/axios.ts
```diff
@@ -20,6 +34,11 @@ axiosInstance.interceptors.request.use(
 
     config.headers['X-Correlation-Id'] = uuidv4()
 
+    // If the request has a payload performs the trim on all its strings
+    if (config.data) {
```
ruggerocastagnola, line 38: Controllare se in questo caso non sia meglio usare `isEmpty` di lodash per un check più profondo (es. oggetto vuoto)

ruggerocastagnola (approved):

PR #504, opened 2023-07-21
Feature: PIN-3621 - Agreement form redesign

alten-dturus:
- Added strings for tooltip and cretifiedAttributes drawer
- Changed general `Drawer` `subtitle` type to accept `React.ReactNode` instead of only `string`
- Set `hideSideNav` to true for `ConsumerAgreementCreatePage`
- Redesigned `ConsumerAgreementCreate` page and implemented the new components


Carminepo2:
We should discuss with @ruggerocastagnola about making the right abstractions on the redesigned agreement components

ruggerocastagnola (approved):

PR #542, opened 2023-09-11
Feature: PIN-3946 - Fix auth server string not being relative to the env

Carminepo2:


ruggerocastagnola:
Closed after a hotfix merge directly on main (see #544)

PR #708, opened 2024-01-31
PIN-4429: Banner archived purpose

diego-dxc:
Shows correct banner when purpose state is archived.


ruggerocastagnola:
Superseded by new features (removed client step in purpose)

PR #707, opened 2024-01-30
Feature: PIN-4391 - Assignment certificate attributes implementation

alten-dturus:
- Updated `generatedTypes`
- Implemented `DialogRevokeCertifiedAttribute`
- Implemented `AssignAttributesTab` and its components and added strings
- Added `AssignAttributesTab` to `TenantCertifierPage`
- Implemented attributes services and hooks and added strings
- Updated strings

PR #712, opened 2024-02-05
Feature: PIN-4391 - Assignment certificate attributes implementation squashed

alten-dturus:


ruggerocastagnola:
We merge this instead of #707

ruggerocastagnola (approved):

PR #724, opened 2024-02-09
Added generate token command

Carminepo2:
This PR adds a new package.json script `npm run generate-token`.
This command allows us to easily retrieve a valid token in order to start the FE locally.

In order to use the script:
- `REMOTE_WELLKNOWN_URL` must be set in the env variables;
- It is required to have a valid aws session with the correct KMS permissions.

Usage:
```bash
npm run generate-token [tenant] [role]
```

[tenant] must be one of the following: `GSP`,`PA1`, `PA2`, `PRIVATO`. Case-insensitive. Default is `GSP`.
[role] must be one of the following: `ADMIN`,`API`, `SECURITY`, `API,SECURITY`, `SUPPORT`. Case-insensitive. Default is `ADMIN`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "interop-dashboard-frontend",
-  "version": "1.0.40",
+  "version": "1.0.45",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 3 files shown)


Carminepo2:
Closing the PR. This script will be implemented in an utility repo someday

PR #766, opened 2024-03-22
Fixed handling the current product in the PagoPA product selection dropdown

ruggerocastagnola:
This PR could have been done in several ways. I chose to have an object so that every time a new environment is created, the frontend is forced to update its config as well. This way, we are forced to set all the `STAGE` values that are allowed. 

`STAGE` values set to `QA` and `DEV` are actually irrelevant, they are there just to avoid breaking the UI, and a comment was put there to remind us

### DIFF
--- a/src/config/env.ts
+++ b/src/config/env.ts
@@ -64,7 +64,15 @@ export const SELFCARE_BASE_URL =
 
 export const STAGE = PAGOPA_ENV?.STAGE ?? 'DEV'
 
-export const SELFCARE_INTEROP_PROD_ID = `prod-interop${STAGE === 'UAT' ? '-coll' : ''}`
+export const SELFCARE_INTEROP_PROD_ID = {
+  UAT: 'prod-interop-coll',
+  PROD: 'prod-interop',
+  ATT: 'prod-interop-atst',
+  // DEV and QA are actually irrelevant. They are set to "prod-interop"
+  // just to avoid breaking the product dropdown in the UI
+  DEV: 'prod-interop',
+  QA: 'prod-interop',
+}[STAGE]
 
 export const PRODUCER_ALLOWED_ORIGINS = PAGOPA_ENV?.PRODUCER_ALLOWED_ORIGINS.split(',')
   .map((o) => o.trim())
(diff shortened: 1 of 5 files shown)


src/config/env.ts
```diff
@@ -64,7 +64,15 @@ export const SELFCARE_BASE_URL =
...
-export const SELFCARE_INTEROP_PROD_ID = `prod-interop${STAGE === 'UAT' ? '-coll' : ''}`
+export const SELFCARE_INTEROP_PROD_ID = {
+  UAT: 'prod-interop-coll',
+  PROD: 'prod-interop',
+  ATT: 'prod-interop-atst',
+  // DEV and QA are actually irrelevant. They are set to "prod-interop"
+  // just to avoid breaking the product dropdown in the UI
+  DEV: 'prod-interop',
+  QA: 'prod-interop',
+}[STAGE]
```
Carminepo2, line 75: I would also throw an error or at least a console warn if the `SELFCARE_INTEROP_PROD_ID` turns out `undefined`

Carminepo2 (changes requested):

ruggerocastagnola:
@Carminepo2 would this be better?

src/utils/common.utils.ts
```diff
@@ -155,3 +156,24 @@ export function clearExponentialInterval(instanceId: string | undefined) {
     exponentialIntervalInstances.delete(instanceId)
   }
 }
+
+export function getCurrentSelfCareProduct() {
```
Carminepo2, line 160:
```suggestion
export function getCurrentSelfCareProductId() {
```

src/config/env.ts
```diff
@@ -64,7 +65,7 @@ export const SELFCARE_BASE_URL =
 
 export const STAGE = PAGOPA_ENV?.STAGE ?? 'DEV'
 
-export const SELFCARE_INTEROP_PROD_ID = `prod-interop${STAGE === 'UAT' ? '-coll' : ''}`
+export const SELFCARE_INTEROP_PROD_ID = getCurrentSelfCareProduct()
```
Carminepo2, line 68: I would remove it from the file at this point, and in case we need it I would directly call the function

Carminepo2: What do you think?

Carminepo2 (changes requested):

Carminepo2 (approved):

PR #880, opened 2024-07-18
Feature/pin 5078 show correlation id for support requests

martinaCampoli:


src/components/layout/ToastNotification.tsx
```diff
@@ -1,32 +1,75 @@
...
+  }
+
   return (
     <Snackbar
       open={isShown}
-      onClose={hideToast}
       autoHideDuration={5000}
       anchorOrigin={{ vertical: 'bottom', horizontal: 'right' }}
-      sx={{ maxWidth: 480 }}
+      sx={{ maxWidth: 350 }}
```
Carminepo2, line 61: Any reason why the `maxWidth` is reduced?

Carminepo2 (changes requested):

src/components/layout/ToastNotification.tsx
martinaCampoli: > Any reason why the `maxWidth` is reduced?

It's reduced to allow the snackbar to remain inside the drawer as in the design

PR #903, opened 2024-07-30
Feature/pin 5128 update mui italia

ruggerocastagnola:


ruggerocastagnola:
Closed because our snapshot tests made it impossible to do

PR #965, opened 2024-10-07
feature/pin-5402-revoked-verified-attribute-update

martinaCampoli:
BFF node: fix revokedBy entry being removed when re-verifying an attribute.
Updated FE: deleted double check on isInVerifiedBy and isInRevokedBy  in isAttributeRevoked function for verified attribute

### DIFF
--- a/src/utils/attribute.utils.ts
+++ b/src/utils/attribute.utils.ts
@@ -47,12 +47,8 @@ export function isAttributeRevoked(
 
       /*
        * The attribute is considered revoked if it has been revoked at least once by any verifier.
-       *
        */
-
-      return typedAttribute.revokedBy.some(({ id }) => {
-        return typedAttribute.revokedBy.some((verifier) => verifier.id === id)
-      })
+      return typedAttribute.revokedBy.length > 0
     case 'declared':
       return Boolean((attribute as DeclaredTenantAttribute).revocationTimestamp)
     default:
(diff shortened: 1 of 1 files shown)


src/utils/attribute.utils.ts
```diff
@@ -36,19 +36,12 @@ export function isAttributeRevoked(
...
 
       if (verifierId) {
-        const isInVerifiedBy = typedAttribute.verifiedBy.some(
-          (verifier) => verifier.id === verifierId
-        )
-
-        if (isInVerifiedBy) return false
         const isInRevokedBy = typedAttribute.revokedBy.some(
           (verifier) => verifier.id === verifierId
         )
```
Carminepo2, line 47: At this point, we could directly return the value without assigning it 

```suggestion
        return typedAttribute.revokedBy.some(
          (verifier) => verifier.id === verifierId
        )
```

Carminepo2 (changes requested):
Also the logic below, outside the `if` should be modified. 
I also remove the `Map`, it's useless now

src/utils/attribute.utils.ts
```diff
@@ -36,38 +36,22 @@ export function isAttributeRevoked(
...
-       * The attribute, in this case, is considered revoked if it is in 'revokedBy' and not in 'verifiedBy'
        */
-      const alreadyCheckedVerifierIds = new Map<string, boolean>()
 
       return typedAttribute.revokedBy.some(({ id }) => {
-        if (alreadyCheckedVerifierIds.has(id)) return false
-        alreadyCheckedVerifierIds.set(id, true)
-        return !typedAttribute.verifiedBy.some((verifier) => verifier.id === id)
+        return typedAttribute.revokedBy.some((verifier) => verifier.id === id)
       })
```
Carminepo2, line 55: The whole block here could be reduced in a simpler check: 
```typescript
return typedAttribute.revokedBy.length > 0
```

Carminepo2 (changes requested):

Carminepo2 (approved):

PR #1014, opened 2024-11-28
feature/pin-5704-delegate-handles-delegator-agreement

martinaCampoli:
- Added delegated chip in eservice cell in providerAgreementTableRow
- Added logic in useGetAgreementsAction for delegator actions


src/pages/ProviderAgreementsListPage/components/ProviderAgreementsTableRow.tsx
```diff
@@ -27,14 +27,26 @@ export const ProviderAgreementsTableRow: React.FC<{ agreement: AgreementListEntr
   const eservice = agreement.eservice
   const descriptor = agreement.descriptor
 
+  const isDelegatedEservice = eservice.producer.id != AuthHooks.useJwt().jwt?.organizationId
```
Carminepo2, line 30:
```suggestion
  const isDelegatedEservice = eservice.producer.id !== AuthHooks.useJwt().jwt?.organizationId
```

Carminepo2 (changes requested):
There is one more thing! 
The `useGetAgreementsActions` hook should be updated following this logic:
Once a delegator delegates the e-service, they could no longer activate or suspend the agreements associated to that e-service, only the delegate from that point on can do that.

PR #1055, opened 2025-01-02
Feature: PIN-5825 - accept and reject delegation

alten-dturus:
- Implemented new services and mutations
- Update `DelegationsReceivedTab`X
- Update dialogs
- Update `useGetDelegationActions`

### DIFF
--- a/src/hooks/useGetDelegationActions.ts
+++ b/src/hooks/useGetDelegationActions.ts
@@ -9,7 +9,7 @@ import { AuthHooks } from '@/api/auth'
 export function useGetDelegationActions(delegation: Delegation | CompactDelegation | undefined) {
   const { t: tCommon } = useTranslation('common', { keyPrefix: 'actions' })
   const { openDialog } = useDialog()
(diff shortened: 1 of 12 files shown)


src/hooks/useGetDelegationActions.ts
```diff
@@ -53,9 +61,9 @@ export function useGetDelegationActions(delegation: Delegation | CompactDelegati
   }
 
   if (
-    delegation.kind === 'DELEGATED_PRODUCER' &&
     delegation.state === 'WAITING_FOR_APPROVAL' &&
-    delegation.delegate.id === jwt?.organizationId
+    delegation.delegate.id === jwt?.organizationId &&
+    isAdmin
```
Carminepo2, line 66: Let's make this if more obvious to read. I would extract `delegation.delegate.id === jwt?.organizationId` into a `isActualTenantDelegate` (or something like that) constant

Carminepo2 (approved):
Just one minor change

PR #1151, opened 2025-03-05
Feature/pin 6330 using tenants table pagination

martinaCampoli:
- Fix: pagination and query params in using tenants tab
- Fix: routes name
- Added strings in mutation-feeback.json


src/api/template/template.queries.ts
```diff
@@ -27,11 +28,11 @@ function getSingleByEServiceTemplateId(eserviceTemplateId: string) {
   })
 }
 
-function getProviderTemplateInstancesList(eServiceTemplateId: string) {
+function getProviderTemplateInstancesList(params: GetEServiceTemplateInstancesParams) {
   //TODO aggiungi params
```
borgesis95, line 32: removing TODO (it's not necessary anymore)

borgesis95 (commented):
good job! just one (stupid) comment

PR #1245, opened 2025-05-13
PIN-6651: New UI for sidenav component

borgesis95:


src/components/sidebar/SidebarItem.tsx
```diff
@@ -0,0 +1,80 @@
...
+    <ListItem sx={{ p: 0 }}>
+      <ListItemButton
+        component={Link}
+        to={to}
+        sx={{
+          ...(isSelected && styles.itemButtonActive),
+        }}
+        selected={isSelected}
+        disabled={disabled}
+        {...props}
```
Carminepo2, line 58: It seems you left the polymorphic aspect in the prop typing but in the implementation you left it out?

src/components/layout/AppLayout.tsx
```diff
@@ -10,6 +11,10 @@ type AppLayoutProps = {
 }
 
 export const AppLayout: React.FC<AppLayoutProps> = ({ children, hideSideNav, sx }) => {
+  const theme = useTheme()
+  const matchMobile = useMediaQuery(theme.breakpoints.down('sm'))
+  const interopRoutes = useGetSidebarItems()
```
Carminepo2, line 16: Let's avoid having a logic based on jwt in this part of the `AppLayout` which is rendered also on public pages

Carminepo2 (approved):

PR #1258, opened 2025-05-16
feat: map backend error codes to localized frontend messages (PIN-10652)

borgesis95:
## 🔗 Issue

[PIN-10652](https://pagopa.atlassian.net/browse/PIN-10652)

## 📝 Description / Context

This PR adds frontend mappings for backend error codes returned by the BFF and process services, so mutation errors can show more specific localized messages instead of only the operation-level fallback.

It also adds a lightweight check to compare mapped frontend errors with backend error definitions and locale files.

## 🛠 List of changes

- Added backend/BFF error-code mappings in `src/config/errors.ts`
- Added localized English and Italian messages for mapped errors
- Updated mutation error handling to show mapped toast messages when available
- Preserved operation-specific fallback labels when a mapped translation is missing
- Added tests for mapped errors and mutation error toast behavior
- Added a pre-push warning check for missing mappings/translations

## 🧪 How to test

- Trigger a mutation that returns a mapped backend error code and verify that the toast shows the specific localized message.
- Trigger a mutation with an unmapped error code and verify that the existing operation-specific fallback message is still shown.
- Run `pnpm check:error-mappings`

[PIN-10652]: https://pagopa.atlassian.net/browse/PIN-10652?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/config/errors.ts
```diff
@@ -0,0 +1,60 @@
...
+    "0035": i18n.t('error:errors-bff.eServiceTemplateInterfaceNotFound'),
+    "0036": i18n.t('error:errors-bff.eServiceTemplateInterfaceDataNotValid'),
+    "0037": i18n.t('error:errors-bff.descriptorTemplateVersionNotFound'),
+} as const;
+
+
+export const getMappedError = (errorCode: string) => {
+    const { service, error } = splitServiceCodeAndMessageCode(errorCode)
+    console.log("service,error", service, typeof error)
+    console.log("err2", errorMessages[error])
```
borgesis95, line 48:
```suggestion
```

src/config/errors.ts
```diff
@@ -0,0 +1,475 @@
...
+  const translationKey = getTranslationKey(errorCode)
+  if (!translationKey) return undefined
+
+  return getLocalizedError(translationKey)
+}
+
+const getTranslationKey = (errorCode?: string): string | undefined => {
+  if (!errorCode) return undefined
+
+  const exactTranslationKey = exactErrorCodes[errorCode]
```
Alepazz, line 459: The lookup matches error codes exactly in the `NNN-NNNN` shape, but the codebase already carries process codes in a 3-digit-suffix shape too: `DUPLICATE_ESERVICENAME_ERROR_CODE` is `'001-007'` (and is compared against real responses in the eservice-template and delegation mutations), while purpose-template codes use the 4-digit `'015-0014'` form. Any process error that arrives 3-digit will miss this map and silently fall back to the generic per-operation label. Could you confirm the actual suffix width returned by the BFF and align the map (or normalize the code before lookup) accordingly?

src/config/query-client.ts
```diff
@@ -186,14 +187,16 @@ mutationCache.config.onError = (error, variables, context, mutation) => {
   if (meta.errorToastLabel) {
     let correlationId
     let errorCode
+    let errorMessage = meta.errorToastLabel
     if (error instanceof AxiosError) {
       correlationId = error.response?.data?.correlationId
       errorCode = error.response?.data?.errors?.[0]?.code
+      errorMessage = getMappedError(errorCode) ?? errorMessage
```
Alepazz, line 194: This is the path that actually changes user-facing behavior, but it isn't covered by tests: the suite for this file only exercises confirmation dialogs, and the new tests check `getMappedError` in isolation. Could you add a case that drives an Axios error response through here and asserts the mapped toast is shown, the correlation data is preserved, and an unmapped code falls back to the original label?

src/config/errors.ts
```diff
@@ -0,0 +1,475 @@
...
+  return undefined
+}
+
+const getLocalizedError = (translationKey: string): string => {
+  const normalizedTranslationKey = translationKeyAliases[translationKey] ?? translationKey
+  const localizedMessage = i18n.t(`error:errors-bff.${normalizedTranslationKey}`, {
+    defaultValue: '',
+  })
+
+  return localizedMessage || i18n.t('error:errors-bff.genericMappedError')
```
Alepazz: Since this always returns a truthy string (the message or `genericMappedError`), `getMappedError` is never `undefined` for a mapped code, so the `?? errorMessage` fallback in the mutation error handler can no longer reach the operation-specific label. For a code that is mapped but whose translation key is missing or misaligned, the user sees the vague generic message instead of the more specific label that shipped before. Returning `undefined` when the translation resolves empty would let the contextual label win.

.husky/pre-push
```diff
@@ -54,6 +54,11 @@ if [ -t 0 ]; then
   exit 0
 fi
 
+log "Checking frontend/backend error mapping alignment..."
+if ! pnpm check:error-mappings; then
```
Alepazz, line 58: `check:error-mappings` runs without `--strict`, so it returns success even when mappings or translations are missing, which means this guard never triggers and every push still pays for the network fetches to the backend repo. Consider gating this behind an opt-in or moving it to CI so a normal push doesn't depend on `raw.githubusercontent.com`.

src/static/locales/it/error.json
```diff
@@ -30,6 +30,313 @@
...
+    "openapiVersionNotRecognized": "La versione OpenAPI non è supportata",
+    "interfaceExtractingInfoError": "Impossibile leggere le informazioni del file di interfaccia",
+    "invalidContentTypeDetected": "Il formato del file caricato non è supportato",
+    "tokenVerificationFailed": "La sessione non è valida",
+    "featureFlagNotEnabled": "Questa funzionalità non è abilitata",
+    "invalidServerUrl": "L'URL del server nel file di interfaccia non è valido",
+    "eServiceDescriptorNotFound": "Descrittore dell'eService non trovato",
+    "eServiceDescriptorWithoutInterface": "Il descrittore dell'eService non ha un'interfaccia",
+    "eServiceDocumentNotFound": "Documento dell'eService non trovato",
+    "draftDescriptorAlreadyExists": "Esiste gia un descrittore in bozza",
```
Alepazz: A few strings in this block are missing their Italian accents, for example `draftDescriptorAlreadyExists` and `interfaceAlreadyExists` (`gia` should be `già`), and several `eservice...` entries where `e` should be `è` and `modalita` should be `modalità`. Worth a quick pass with whoever owns the copy.

Alepazz (changes requested):
A few points to sort out before merge, see inline comments.

.husky/pre-push
DenisLantero: This is the intentional behavior because if we run it with `--strict` or if we add the check in CI it could happen that we see an error about something that we didn't touch, making it a bit confusing.

Imagine if the backend adds an error, the script would detect it and when you push / in CI it would report it as an error, but it might not have anything to do with the current PR / commit.

I decided to add it as a warning in the pre-push script so this way we are at least aware when there is a missing error translation, but we are not forced to fix it immediatly / in the current branch.

src/config/errors.ts
DenisLantero: Are you sure that the backend uses 3-digits suffixes for error codes?

I only found 2 errors that used that shape:

```
catalog-process: eServiceNameDuplicateForProducer: "007"  -> 001-007
tenant-process: expirationDateCannotBeInThePast: "009"   -> 005-009
```

It seems to me like a backend definiton problem, do you agree?

Anyways, I added these 2 missing errors to the maps [here](https://github.com/pagopa/pdnd-interop-frontend/pull/1258/commits/9d3664c76508c402c0b6f5bf678b3286a234fe20)

src/config/query-client.ts
DenisLantero: Added [here](https://github.com/pagopa/pdnd-interop-frontend/pull/1258/commits/6f1caff1ce8caaafa3a84e2d1e80f4917d309082)

src/static/locales/it/error.json
DenisLantero: Fixed [here](https://github.com/pagopa/pdnd-interop-frontend/pull/1258/commits/ee6a5bbd1666842efd8358b8e6a7c73c00cd134f)

src/config/errors.ts
DenisLantero: Fixed [here](https://github.com/pagopa/pdnd-interop-frontend/pull/1258/commits/f9283b319bb3a8fcf1902aafa536399ec8dee980)

Alepazz (approved):
LGTM, the replies are clear.

src/config/errors.ts
Alepazz: Agreed, this looks like a backend definition inconsistency rather than a frontend issue, and adding the two known codes covers the current cases. One thing to keep on the radar: if the backend ever introduces a third 3-digit code, it will silently fall back to the generic label again, since the map is keyed on the 4-digit shape. Worth keeping in mind, and possibly investigating a normalization step later, but fine to leave as is for now.

.husky/pre-push
Alepazz: Makes sense, keeping it as a non-blocking pre-push warning is a reasonable tradeoff to avoid surfacing unrelated backend errors on every push or in CI. Fine to leave as is.

src/config/errors.ts
DenisLantero: I created a [backend issue](https://pagopa.atlassian.net/browse/PIN-10485) for this

PR #1275, opened 2025-05-28
PIN-6879 - Risk analysis exporter tool

Carminepo2:
Closes [PIN-6879](https://pagopa.atlassian.net/browse/PIN-6879)

[PIN-6879]: https://pagopa.atlassian.net/browse/PIN-6879?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/pages/RiskAnalysisExporterToolPage/components/RiskAnalysisExporterToolContext.tsx
+++ b/src/pages/RiskAnalysisExporterToolPage/components/RiskAnalysisExporterToolContext.tsx
@@ -1,0 +1,98 @@
+import React, { useCallback, useMemo, useState } from 'react'
+import { createContext } from '@/utils/common.utils'
+import type { TenantKind } from '@/api/api.generatedTypes'
+import { TenantHooks } from '@/api/tenant'
+import { getRiskAnalysisKind } from '@/utils/risk-analysis-form.utils'
+import type { RiskAnalysisKind } from '@/types/risk-analysis-form.types'
+import noop from 'lodash/noop'
+
+type RiskAnalysisExporterToolState = {
+  selectedRiskAnalysisKind: RiskAnalysisKind
+  errors: Array<string>
+  output: Record<string, string[]>
+}
+
+type RiskAnalysisExporterToolContextProvider = RiskAnalysisExporterToolState & {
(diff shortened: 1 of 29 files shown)


src/pages/RiskAnalysisExporterToolPage/components/RiskAnalysisExporterToolContext.tsx
```diff
@@ -0,0 +1,98 @@
...
+  forward,
+}: {
+  children: React.ReactNode
+  back: VoidFunction
+  forward: VoidFunction
+}) {
+  const { data: currentTenant } = TenantHooks.useGetActiveUserParty()
+  const tenantRiskAnalysisKind = getRiskAnalysisKind(currentTenant.kind as TenantKind)
+
+  const [toolState, setToolState] = useState<RiskAnalysisExporterToolState>({
```
borgesis95, line 51: Could we rename this? this is strict related to risk analysis you've compiled. I'd like to use something like `riskAnalysisState`

src/pages/RiskAnalysisExporterToolPage/components/RiskAnalysisToolFormStep.tsx
```diff
@@ -0,0 +1,118 @@
...
+              {t('generateCodeButton')}
+            </Button>
+          </Stack>
+        </Box>
+      </FormProvider>
+    </Box>
+  )
+}
+
+function RiskAnalysisKindSelectSection() {
```
borgesis95, line 77: Could we generalize this component? i'm quite sure that we'll use this in other pages (https://github.com/pagopa/pdnd-interop-frontend/pull/1253 this is an example)

Carminepo2: I would avoid prematurely generalize a component. I see that the labels are different, plus there is no alert. 
The generalization would be just for the sake on not repeating  Select input code which I think is overkill.
We could end up with yet another file in the shared components

borgesis95 (approved):

PR #1285, opened 2025-06-05
Feature/pin 6874 implement nomination flow sh

martinaCampoli:
- Added sh section provider side
- Added sh section consumer side

### DIFF
--- a/src/config/env.ts
+++ b/src/config/env.ts
@@ -120,3 +120,6 @@ export const API_SIGNAL_HUB_PULL_INTERFACE_URL =
   isProduction && PAGOPA_ENV
     ? PAGOPA_ENV['API_SIGNAL_HUB_PULL_INTERFACE_URL']
     : `https://raw.githubusercontent.com/pagopa/interop-signalhub-core/refs/heads/main/docs/openAPI/pull-signals.yaml`
+
+export const SIGNALHUB_PERSONAL_DATA_PROCESS_URL =
+  isProduction && PAGOPA_ENV ? PAGOPA_ENV['SIGNALHUB_PERSONAL_DATA_PROCESS_URL'] : ``
--- a/src/config/env.ts
+++ b/src/config/env.ts
@@ -123,3 +123,6 @@ export const API_SIGNAL_HUB_PULL_INTERFACE_URL =
 
 export const SIGNALHUB_PERSONAL_DATA_PROCESS_URL =
   isProduction && PAGOPA_ENV ? PAGOPA_ENV['SIGNALHUB_PERSONAL_DATA_PROCESS_URL'] : ``
+
+export const SIGNALHUB_GUIDE_URL =
+  'https://developer.pagopa.it/pdnd-interoperabilita/guides/manuale-operativo-signal-hub'
--- a/src/config/env.ts
+++ b/src/config/env.ts
@@ -123,6 +123,3 @@ export const API_SIGNAL_HUB_PULL_INTERFACE_URL =
 
 export const SIGNALHUB_PERSONAL_DATA_PROCESS_URL =
   isProduction && PAGOPA_ENV ? PAGOPA_ENV['SIGNALHUB_PERSONAL_DATA_PROCESS_URL'] : ``
-
-export const SIGNALHUB_GUIDE_URL =
-  'https://developer.pagopa.it/pdnd-interoperabilita/guides/manuale-operativo-signal-hub'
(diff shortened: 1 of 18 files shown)


src/pages/ProviderEServiceCreatePage/components/EServiceCreateStepGeneral/EServiceCreateStepGeneral.tsx
```diff
@@ -231,39 +234,46 @@ export const EServiceCreateStepGeneral: React.FC = () => {
             sx={{ mb: 0, mt: 3 }}
             onValueChange={(mode) => onEserviceModeChange!(mode as EServiceMode)}
           />
-          {isSignalHubFlagEnabled && (
+        </SectionContainer>
+
+        {isSignalHubFlagEnabled && (
```
borgesis95, line 239: Could we put this on a separeted component ? (it's ok that this new component is in same file)

src/pages/ProviderEServiceDetailsPage/components/ProviderEServiceDetailsTab/ProviderEServiceDetailsTab.tsx
```diff
@@ -24,6 +30,11 @@ export const ProviderEserviceDetailsTab: React.FC = () => {
     EServiceQueries.getDescriptorProvider(eserviceId, descriptorId)
   )
 
+  const producerId = AuthHooks.useJwt().jwt?.organizationId as string
+  const isSignalHubFlagEnabled = FEATURE_FLAG_SIGNALHUB_WHITELIST
```
borgesis95, line 34: This var `isSignalHubFlagEnabled` is present in several files and contain a bit of logic (if flag is active you have to check if user is on whitelist as well). That's fine but i'd like to move this into a file (something like feature-flags utils) in order to have a unique source of truth (in case we need to remove o replace logic)

src/pages/ProviderEServiceDetailsPage/components/ProviderEServiceDetailsTab/ProviderEServiceSignalHubSection.tsx
```diff
@@ -0,0 +1,75 @@
...
+  return (
+    <SectionContainer
+      title={t('title')}
+      description={
+        <Typography color="text.secondary" variant="body2">
+          {t('description.before')}{' '}
+          <IconLink href={''} target="_blank" endIcon={<LaunchIcon fontSize="small" />}>
+            {' '}
+            {/** TODO LINK */}
+            {t('description.linkLabel')}
```
borgesis95, line 35: instead of TODO LINK put this : [soluzione integrata]( https://developer.pagopa.it/pdnd-interoperabilita/guides/manuale-operativo-signal-hub) (of course put it into a config var).

borgesis95 (changes requested):

src/config/env.ts
```diff
@@ -120,3 +120,9 @@ export const API_SIGNAL_HUB_PULL_INTERFACE_URL =
   isProduction && PAGOPA_ENV
     ? PAGOPA_ENV['API_SIGNAL_HUB_PULL_INTERFACE_URL']
     : `https://raw.githubusercontent.com/pagopa/interop-signalhub-core/refs/heads/main/docs/openAPI/pull-signals.yaml`
+
+export const SIGNALHUB_PERSONAL_DATA_PROCESS_URL =
+  isProduction && PAGOPA_ENV ? PAGOPA_ENV['SIGNALHUB_PERSONAL_DATA_PROCESS_URL'] : ``
+
+export const SIGNALHUB_GUIDE_URL =
```
borgesis95, line 127: in this case i'd like to have it on `constants.ts` file  (for `SIGNALHUB_GUIDE_URL`)

src/pages/ProviderEServiceCreatePage/components/EServiceCreateStepGeneral/EServiceCreateStepGeneral.tsx
```diff
@@ -231,40 +236,15 @@ export const EServiceCreateStepGeneral: React.FC = () => {
...
-                sx={{ my: 0 }}
-              />
-            </SectionContainer>
-          )}
         </SectionContainer>
 
+        {isSignalHubFlagEnabled && (
+          <SignalHubSection
+            isEserviceFromTemplate={isEserviceFromTemplate}
+            areEServiceGeneralInfoEditable={areEServiceGeneralInfoEditable}
```
borgesis95, line 244: instead of pass two props, we can pass a generic one which will be sum of both (like `isSignalhubActivationEditable` or similar). That's why this new component doesn't care about specific info (is Eservice is an instance of template or other stuff) but just if is editable or not.

borgesis95 (changes requested):

PR #1293, opened 2025-06-10
feature/pin-6986-edit-flag-sh-esevice-already-existing

martinaCampoli:
- Added UI action editing isSignaHubEnabled
- Added Drawer for editing signal hub section


src/api/eservice/eservice.mutations.ts
```diff
@@ -493,6 +493,20 @@ function useDeleteDraftAndUpgradeEService() {
   })
 }
 
+function useUpdateEServiceIsSignalHubEnabled() {
```
borgesis95, line 496: I'd use useUpdateEServiceIsSignalHub without word enabled (we can't know if user will be enable or disable SH service)

src/api/eservice/eservice.services.ts
```diff
@@ -577,6 +578,17 @@ async function getIsEServiceNameAvailable({ eserviceName }: { eserviceName: stri
   return response.data
 }
 
+async function updateEServiceIsSignalHubEnabled({
```
borgesis95, line 581:
```suggestion
async function updateEServiceSignalhub
```

borgesis95 (changes requested):

PR #1304, opened 2025-06-19
fix/pin-7052-error-editing-eservice-template-draft

martinaCampoli:
Fixed template id in updateDraft call in create step general


src/pages/ProviderEServiceTemplateCreatePage/components/EServiceTemplateCreateStepGeneral/EServiceTemplateCreateStepGeneral.tsx
```diff
@@ -67,7 +67,10 @@ export const EServiceTemplateCreateStepGeneral: React.FC = () => {
       const isEServiceTemplateTheSame = compareObjects(formValues, template.eserviceTemplate)
 
       if (!isEServiceTemplateTheSame)
-        updateDraft({ eServiceTemplateId: template.id, ...formValues }, { onSuccess: forward })
+        updateDraft(
+          { eServiceTemplateId: template.eserviceTemplate.id, ...formValues },
```
Carminepo2, line 71: We should really change the `template` name in something more descriptive, in order to avoid confusion and other bugs like these in the future.
`templateVersion` should be more appropriate

borgesis95 (approved):

PR #1328, opened 2025-07-04
Feature: PIN-7089 - add correlation id to assistance navigation

alten-dturus:
- Implemented correlationId store
- Add correlationId to assistance navigation


src/stores/correlation-id.store.ts
```diff
@@ -0,0 +1,50 @@
...
+  timeoutId: NodeJS.Timeout | null
+  setCorrelationId: (correlationId: string) => void
+  clearCorrelationId: () => void
+}
+
+export const useCorrelationIdStore = create<CorrelationIdStoreType>((set, get) => ({
+  correlationId: null,
+  timeoutId: null,
+  setCorrelationId: (correlationId: string) => {
+    console.log('correlationId setting', Date.now(), correlationId)
```
borgesis95, line 14:
```suggestion
```

src/stores/correlation-id.store.ts
```diff
@@ -0,0 +1,50 @@
...
+    if (currentTimeoutId) {
+      clearTimeout(currentTimeoutId)
+    }
+
+    set({ correlationId })
+
+    const newTimeoutId = setTimeout(() => {
+      get().clearCorrelationId()
+      console.log('correlationId expired', Date.now(), correlationId)
+    }, 10000)
```
borgesis95, line 25: move this on constants file

borgesis95 (changes requested):
Can we add test for Header.tsx ( in this case just to check what is behaviour when correlationId is available and when is not)

borgesis95 (approved):

PR #1477, opened 2025-10-21
feature/PIN-7915-add-filter-in-create-delegation

martinaCampoli:
Added filters to api calls for eservices  and template to show in create eservice in delegation form


src/pages/DelegationCreatePage/components/DelegationCreateEServiceAutocomplete.tsx
```diff
@@ -67,6 +66,7 @@ export const DelegationCreateEServiceAutocomplete: React.FC<
       limit: 50,
       offset: 0,
       delegated: false,
+      personalData: true,
```
borgesis95: has this var to be always `true`?

martinaCampoli: Yes, we want to show the user only eservices/templates that have personal data set. Below is the comment for the filter in the generated types.
`/** if true only e-services that handle personal data will be returned, if false only non-personal data e-services will be returned, if not present all e-services will be returned */`

PR #1478, opened 2025-10-21
PIN-7319 notification badge and mark as read by entity

sandrotaje:
image


borgesis95 (commented):
Have you think about to move this 
```
React.useEffect(() => {

    if (purposeId) {

      markNotificationsAsRead({ entityId: purposeId })

    }

  }, [purposeId, markNotificationsAsRead])
```

and its mutation to custom hook? We copy/paste this in several pages. In this way could be centralized (instead of purposeId we could use entityId as hook's parameter

PR #1609, opened 2026-01-07
feature/PIN-8817: add chance to choose how many element user can see within list

borgesis95:


Alepazz (changes requested):
**Point 3 - Partially resolved**
The custom limits passed to `usePagination()` are now restored (`{ limit: 3 }` for documentation drawers, `{ limit: 12 }` for catalog pages), which fixes the initial limit mismatch. However, the commit also removes the `rowPerPageOptions` with custom options `[12, 24, 36]` from three catalog pages (_ConsumerEServiceCatalog_, _ConsumerPurposeTemplateCatalog_, _ProviderEServiceTemplatesCatalog_). This means the Pagination component will now use its default options. 
Two possible issues I can notice:
  - If the defaults don't include 12, the initial limit won't be selectable in the dropdown (same mismatch problem, reversed)
  - If `rowPerPageOptions` not being passed means the dropdown isn't rendered at all, the user loses the ability to change items per page (regression)

  Could you verify which case applies?

PR #1610, opened 2026-01-08
[bugfix] in app notification preference saved

GorlemZ:
The POST payload was built with the email preference flag because the in-app tab’s handler forwarded its second argument (meant for the email flow) instead of the first. By deconstructing the params correctly, now it saves the correct value in the API payload


borgesis95:
this problem has already been addressed.

PR #1629, opened 2026-01-20
refactor/PIN-8951-refactor-useGetConsumerPurposesActions

martinaCampoli:
## Issue

[PIN-8951](https://pagopa.atlassian.net/browse/PIN-8951)

Refactor logic in useGetConsumerPurposesActions and added some tests

[PIN-8951]: https://pagopa.atlassian.net/browse/PIN-8951?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/hooks/useGetConsumerPurposesActions.ts
+++ b/src/hooks/useGetConsumerPurposesActions.ts
@@ -185,17 +185,8 @@ function useGetConsumerPurposesActions(purpose?: Purpose) {
       {
         hasCurrentVersion: false,
         hasWaitingForApprovalVersion: true,
-        isSuspendedByConsumer: false,
       },
-      () => [deleteAction]
-    )
-    .with(
-      {
-        hasCurrentVersion: false,
-        hasWaitingForApprovalVersion: true,
-        isSuspendedByConsumer: true,
-      },
-      () => []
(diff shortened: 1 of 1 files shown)


src/hooks/useGetConsumerPurposesActions.ts
```diff
@@ -132,52 +133,120 @@ function useGetConsumerPurposesActions(purpose?: Purpose) {
...
-  }
+      if (isSuspended && isSuspendedByConsumer) {
+        actions.push(activateAction)
+      }
 
-  if (isSuspended && isSuspendedByConsumer) {
-    actions.push(activateAction)
-  }
+      return actions
+    })
```
Alepazz: Maybe I'm wrong but I can see differences in business logic comparing before and after refactoring.

The original code returned an empty array when `!purpose.currentVersion && purpose.waitingForApprovalVersion && purpose.suspendedByConsumer`:
`return { actions: purpose.suspendedByConsumer ? [] : [deleteAction] }`

The new pattern only handles `isSuspendedByConsumer: false`. When it's `true`, the code falls through this `otherwise()` which returns actions instead of `[]`.

Am I missing something?

borgesis95 (approved):
good job!

PR #1632, opened 2026-01-21
feature: lang integration from selfcare (PIN-7690) 

alten-dturus:


src/components/sidebar/InteropSidebarItems.tsx
```diff
@@ -28,13 +29,18 @@ export const InteropSidebarItems: React.FC<InteropSidebarItems> = ({ routes }) =
...
   const pathname = useCurrentRoute().routeKey
   const { jwt, isAdmin, isSupport } = AuthHooks.useJwt()
 
   const selfcareUsersPageUrl =
-    jwt && `${SELFCARE_BASE_URL}/dashboard/${jwt.selfcareId}/users#${getCurrentSelfCareProductId()}`
-  const selfcareGroupsPageUrl = jwt && `${SELFCARE_BASE_URL}/dashboard/${jwt.selfcareId}/groups`
+    jwt &&
+    `${SELFCARE_BASE_URL}/dashboard/${
+      jwt.selfcareId
+    }/users#${getCurrentSelfCareProductId()}?lang=${lang}`
```
Alepazz: The `?lang=` query parameter is appended after the `#` fragment, which means it becomes part of the fragment identifier and is never interpreted as an actual query parameter.
[Here](https://datatracker.ietf.org/doc/html/rfc3986#section-3) the doc about that.

PR #1653, opened 2026-02-06
PIN-9106 adjust new eservice step one

antofo:
- Add required fields UI indicator
- Adjust step sections

### DIFF
--- a/src/pages/ProviderEServiceFromTemplateCreatePage/ProviderEServiceFromTemplateCreate.page.tsx
+++ b/src/pages/ProviderEServiceFromTemplateCreatePage/ProviderEServiceFromTemplateCreate.page.tsx
@@ -120,7 +120,7 @@ const ProviderEServiceFromTemplateCreate: React.FC = () => {
       }
       backToAction={{
         label: t('backToListBtn'),
-        to: 'PROVIDE_ESERVICE_TEMPLATE_CATALOG',
+        to: 'PROVIDE_ESERVICE_LIST',
       }}
     >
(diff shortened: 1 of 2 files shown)


borgesis95 (commented):
Can i ask to add test for `EServiceCreateStepGeneral.tsx` and `ProviderEserviceCreate.page.tsx`? 

I know that are component that was already there, but should be a good idea add tests for this component in order to make our code more resilient.

For example, for EServiceCreateStepGeneral.tsx i would like to test (this is just an idea):
- what happen if is e-service come from e-service template
- navigation to correct route if everything goes fine and other 
- what happen if personalData is `undefined`
etc...

src/pages/ProviderEServiceCreatePage/components/EServiceCreateStepGeneral/EServiceCreateStepGeneral.tsx
```diff
@@ -202,8 +198,15 @@ export const EServiceCreateStepGeneral: React.FC = () => {
...
             sx={{ mb: 0, mt: 3 }}
+            required
           />
-
+        </SectionContainer>
+        <SectionContainer title={t('create.step1.detailsSection.title')}>
+          {!isEserviceFromTemplate && (
+            <Alert severity="warning" sx={{ mb: 0, mt: 3 }}>
+              {t('create.step1.firstVersionOnlyEditableInfo')}
+            </Alert>
```
Alepazz: Did you test the UI appearance of this alert? The original alert had the `mb: 3` that now is not present.
I don't know, I'm just asking about the UI experience.

PR #1706, opened 2026-02-25
feat: instanceLabel edit after publication (PIN-9266)

GorlemZ:
## Summary
- Add instanceLabel edit drawer on published e-service detail page with validation and duplicate error handling
- Add instanceLabel read-only display in pre-publication summary page (conditional on template e-services)
- Improve detail section: show "-" placeholder with secondary color when instanceLabel is undefined
- Fix drawer catalog preview layout (stacked instead of inline)
- Add tests for UpdateInstanceLabelDrawer, ProviderEServiceGeneralInfoSection and ProviderEServiceGeneralInfoSummary

### DIFF
--- a/src/components/shared/UpdateInstanceLabelDrawer.tsx
+++ b/src/components/shared/UpdateInstanceLabelDrawer.tsx
@@ -34,11 +34,9 @@ export const UpdateInstanceLabelDrawer = React.forwardRef<
     keyPrefix: 'read.drawers.updateInstanceLabelDrawer',
   })
 
-  const defaultValues = {
-    instanceLabel: currentInstanceLabel,
-  }
-
-  const formMethods = useForm<UpdateInstanceLabelFormValues>({ defaultValues })
+  const formMethods = useForm<UpdateInstanceLabelFormValues>({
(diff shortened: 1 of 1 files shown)


src/components/shared/UpdateInstanceLabelDrawer.tsx
```diff
@@ -0,0 +1,114 @@
...
+    keyPrefix: 'read.drawers.updateInstanceLabelDrawer',
+  })
+
+  const defaultValues = {
+    instanceLabel: currentInstanceLabel,
+  }
+
+  const formMethods = useForm<UpdateInstanceLabelFormValues>({ defaultValues })
+
+  const watchedInstanceLabel = formMethods.watch('instanceLabel')
```
Alepazz: The `defaultValues` variable is defined at render time and used in `handleTransitionExited` as a closure. If the drawer close animation completes **before** React re-renders with the updated `currentInstanceLabel`, the
  reset could use a stale value. A simple fix is to call `formMethods.reset()` with no arguments, which uses the internal defaults already updated by the `useEffect`.

PR #1712, opened 2026-02-26
feat: reduce template name maxLength from 60 to 45 (PIN-9286)

GorlemZ:
## Summary
- Reduce template name max length from 60 to 45 characters in creation step, edit drawer, and related i18n labels
- Make `maxLength` a configurable prop on `UpdateNameDrawer` (default 60, overridden to 45 for templates)


borgesis95 (commented):
is it worth put 45 into a `constants.ts`? In order to have unique source of truth

PR #1714, opened 2026-02-26
feat: consumer and producer purpose view page (PIN-9196)

l-giacomini:
## Issue
- [PIN-9196](https://pagopa.atlassian.net/browse/PIN-9196)
- Changed copy of consumer purpose view page
- Changed copy of producer purpose view page
- Removed alertExceedingThreshold info alert
- Fixed spacings

[PIN-9196]: https://pagopa.atlassian.net/browse/PIN-9196?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/__mocks__/data/purpose.mocks.ts
+++ b/__mocks__/data/purpose.mocks.ts
@@ -215,7 +215,12 @@ const createMockPurposeCallsExceed = createMockFactory<Purpose>({
       audience: ['test'],
     },
   },
-  agreement: { id: 'agreement-id', state: 'ACTIVE', canBeUpgraded: false, consumerId: 'consumer-id' },
+  agreement: {
+    id: 'agreement-id',
+    state: 'ACTIVE',
+    canBeUpgraded: false,
+    consumerId: 'consumer-id',
+  },
   riskAnalysisForm: {
     answers: { usesPersonalData: ['YES'] },
     version: '3.1',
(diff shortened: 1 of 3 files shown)


src/components/shared/IconLink.tsx
```diff
@@ -50,6 +50,7 @@ export const IconLink = <D extends React.ElementType = 'a'>({
             cursor: 'pointer',
             textDecoration: 'inherit',
             display: 'inline-flex',
+            fontWeight: 700,
```
Alepazz: Adding `fontWeight: 700` here affects **all** usages of `IconLink` across the app, not just the purpose detail pages. Is this intentional? If only the purpose links need bold text, it might be better to pass it via the `sx` prop at the call site rather than hardcoding it in the shared component.

src/pages/ConsumerPurposeDetailsPage/components/PurposeDetailsTab/ConsumerPurposeDetailsGeneralInfoSection.tsx
```diff
@@ -104,12 +104,17 @@ export const ConsumerPurposeDetailsGeneralInfoSection: React.FC<
...
+              content={purpose.consumer.name}
+            />
+            <InformationContainer
+              label={t('delegatedConsumerField.label')}
+              content={purpose.delegation.delegate.name}
+            />
+          </>
         )}
         {isFromPurposeTemplate && (
           <InformationContainer
```
Alepazz: I agree with @borgesis95's question here. Before this change, the consumer name was always visible. Now it's hidden when there's no delegation, which seems like a regression — non-delegated purposes should still display who the consumer is.

The new test at line 193 also asserts `consumerField.label` is `not.toBeInTheDocument()` when delegation is missing, which confirms this is intentional — but is it correct from a product standpoint?

Alepazz (commented):
Good work on the copy updates and test additions. I agree with @borgesis95's comment about the `purpose.consumer.name` field being moved inside the delegation check — see my inline comment on that.

I have a few additional observations below.

src/pages/ConsumerPurposeDetailsPage/components/PurposeDetailsTab/ConsumerPurposeDetailsGeneralInfoSection.tsx
l-giacomini: Here is the comment by the design team:

https://www.figma.com/design/CpRV3kPvFEWLXGtJUgWeZW/Interop-%E2%80%94-Delivery-FE---QA?node-id=1450-44660&t=BYfd5uZP5yNOt1XP-4

PR #1722, opened 2026-03-02
fix: move variable product id to env to fix wrong product (PIN-9283)

alten-dturus:
## 🔗 Issue

[PIN-9283](https://pagopa.atlassian.net/browse/PIN-9283)
There was a problem with the function that did the map from STAGE to productId resulting in a wrong redirect when changing the product.

## 📝 Description / Context

We removed the function and moved the productId to env file so like that we get the right productId from env and we hide an information that will be clear in code otherwise

[PIN-9283]: https://pagopa.atlassian.net/browse/PIN-9283?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/config/constants.ts
+++ b/src/config/constants.ts
@@ -17,7 +17,7 @@ export const pagoPaLink = {
   ariaLabel: 'Vai al sito di PagoPA S.p.A.',
   title: 'Vai al sito di PagoPA S.p.A.',
 }
-export const assistanceLink = `https://selfcare.pagopa.it/assistenza?productId=${getCurrentSelfCareProductId()}`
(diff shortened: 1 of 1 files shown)


Alepazz (approved):
LGTM! One minor thing: the `import { SELFCARE_PRODUCT_ID } from '@/config/env'` added in `setupTests.ts` is unused — the `testConfigs` object uses the hardcoded string `'prod-interop'` instead of the imported variable. The import should be removed.

Also, the PR description is missing the PIN reference. It should include [PIN-9283](https://pagopa.atlassian.net/browse/PIN-9283).

PR #1752, opened 2026-03-06
feature: adjust upload interface step on e-service create (PIN-9390)

antofo:
## Jira Issue
[PIN-9390](https://pagopa.atlassian.net/browse/PIN-9390)

## Context/Why
**Upload interface error**
- Adjust validation for upload interface section

## Key Changes
 - Add upload interface error state in *EServiceCreateStepTechSpec*
 

[PIN-9390]: https://pagopa.atlassian.net/browse/PIN-9390?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/pages/ProviderEServiceCreatePage/components/EServiceCreateStepTechSpec/EServiceCreateStepTechSpec.tsx
+++ b/src/pages/ProviderEServiceCreatePage/components/EServiceCreateStepTechSpec/EServiceCreateStepTechSpec.tsx
@@ -4,7 +4,7 @@ import { StepActions } from '@/components/shared/StepActions'
 import type { ActiveStepProps } from '@/hooks/useActiveStep'
 import { minutesToSeconds, secondsToMinutes } from '@/utils/format.utils'
 import { Box } from '@mui/material'
-import React, { useEffect, useState } from 'react'
+import React, { useRef } from 'react'
 import { FormProvider, type SubmitHandler, useForm } from 'react-hook-form'
 import { useTranslation } from 'react-i18next'
 import { useEServiceCreateContext } from '../EServiceCreateContext'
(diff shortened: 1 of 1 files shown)


src/pages/ProviderEServiceCreatePage/components/EServiceCreateStepTechSpec/EServiceCreateStepTechSpec.tsx
```diff
@@ -107,11 +113,18 @@ export const EServiceCreateStepTechSpec: React.FC<ActiveStepProps> = () => {
       </>
     )
 
+  useEffect(() => {
+    if (descriptor?.interface) {
+      setInterfaceError(undefined)
+    }
+  }, [descriptor?.interface])
```
Alepazz: This `useState` + `useEffect` combo is used to derive state that could be computed directly. The React docs recommend avoiding effects for derived state: https://react.dev/learn/you-might-not-need-an-effect#updating-state-based-on-props-or-state

A simpler approach using a ref to track whether the user has attempted to submit:

```ts
const hasSubmitted = useRef(false)

const interfaceError = hasSubmitted.current && !descriptor?.interface
  ? t('step4.interface.requiredError')
  : undefined

const onSubmit: SubmitHandler<EServiceCreateStepTechSpecFormValues> = (values) => {
  if (!descriptor) return
  hasSubmitted.current = true
  if (!descriptor.interface) return
  // ...
}
```

This removes both the `useState` and the `useEffect`, and the error naturally clears when `descriptor.interface` becomes truthy (on the next render after upload).

Alepazz (approved):

PR #1800, opened 2026-04-01
feature: change fragments management on app init (PIN-9631)

antofo:
## Jira Issue
[PIN-9631](https://pagopa.atlassian.net/browse/PIN-9631)

## Context/Why
**Redirect on app initialization**
- Change fragments management on app startup

## Key Changes
 - Change fragments management in `App.tsx`
 - Add `lang` parameter defaulted to `i18n` current selected language

[PIN-9631]: https://pagopa.atlassian.net/browse/PIN-9631?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -24,9 +24,11 @@ const urlParams = new URLSearchParams(window.location.search)
 const redirectUrl = urlParams.get('redirectUrl')
 
 if (redirectUrl) {
-  const hashes = window.location.hash.replace('#', '').split('&')
-  const selfCareIdentityToken = hashes.find((hash) => hash.includes('id='))?.split('=')[1] ?? ''
-  const lang = hashes.find((hash) => hash.includes('lang='))?.split('=')[1] ?? i18n.language
+  const fragments = window.location.hash.replace('#', '').split('&')
+  const selfCareIdentityToken =
+    fragments.find((fragment) => fragment.includes('id='))?.split('=')[1] ?? ''
+  const lang =
+    fragments.find((fragment) => fragment.includes('lang='))?.split('=')[1] ?? i18n.language
   const url = `/ui/${lang}${redirectUrl}#id=${selfCareIdentityToken}`
   window.location.replace(url)
 } else {
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -25,7 +25,6 @@ const urlParams = new URLSearchParams(window.location.search)
 const redirectUrl = urlParams.get('redirectUrl')
 
 const fragmentParams = new URLSearchParams(window.location.hash.replace('#', ''))
-const selfCareIdentityToken = fragmentParams.get('id') ?? ''
 const requestedLang = fragmentParams.get('lang')?.toLowerCase() ?? ''
 const isSupportedLang = requestedLang in LANGUAGES
 const lang = isSupportedLang ? requestedLang : DEFAULT_LANG
@@ -35,12 +34,13 @@ if (requestedLang) {
 }
 
 if (redirectUrl) {
-  const url = `/ui/${i18n.language}${redirectUrl}#id=${selfCareIdentityToken}`
+  const selfCareIdentityToken = fragmentParams.get('id') ?? ''
+  const url = `/ui/${lang}${redirectUrl}#id=${selfCareIdentityToken}`
 
   window.location.replace(url)
 } else if (requestedLang) {
   fragmentParams.delete('lang')
-  const url = `/ui/${i18n.language}/catalogo-e-service#${fragmentParams.toString()}`
+  const url = `/ui/${lang}/catalogo-e-service#${fragmentParams.toString()}`
 
   window.location.replace(url)
 } else {
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -31,7 +31,6 @@ if (lang) {
   i18n.changeLanguage(lang)
 }
 
-console.log('LANG', lang)
 if (redirectUrl) {
   const url = `/ui/${i18n.language}${redirectUrl}#id=${selfCareIdentityToken}`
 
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -24,8 +24,12 @@ const urlParams = new URLSearchParams(window.location.search)
 const redirectUrl = urlParams.get('redirectUrl')
 
 if (redirectUrl) {
-  const selfCareIdentityToken = window.location.hash.replace('#id=', '')
-  const url = `/ui/${i18n.language}${redirectUrl}#id=${selfCareIdentityToken}`
+  const fragments = window.location.hash.replace('#', '').split('&')
+  const selfCareIdentityToken =
```
Alepazz: `includes('id=')` will match any fragment containing the substring `id=` (e.g., `sessionId=abc`). Same for `lang=` (e.g., `slang=test`). Consider using `startsWith('id=')` / `startsWith('lang=')` for an exact parameter match.

Alepazz: `.split('=')[1]` will truncate the value if the token contains `=` characters (common in base64-encoded tokens). The previous code (`window.location.hash.replace('#id=', '')`) did not have this issue. Consider using `.split('=').slice(1).join('=')` to preserve `=` in the value.

antofo: resolved

antofo: resolved

Alepazz (approved):

borgesis95 (changes requested):
I think that in `dev` env (and other enviroments) this can't work. Here we check if lang is available (and handle it) only if there is `redirectUrl` (this happen only when user click from an email notification). We need to handle this for each case

src/api/auth/auth.services.ts
```diff
@@ -46,10 +55,21 @@ async function getSessionToken(): Promise<string | null> {
 
   // 3. See if we are trying to login as support operator
   // If the url has contains saml2 and jwt, we are trying to login as support operator
-  const hasSupportOperatorToken =
-    window.location.hash.includes('#saml2=') && window.location.hash.includes('jwt=')
+  const saml2 =
```
borgesis95: what is this?

borgesis95 (approved):

src/App.tsx
```diff
@@ -17,15 +17,31 @@ import { queryClient } from './config/query-client'
...
+const isSupportedLang = requestedLang in LANGUAGES
+const lang = isSupportedLang ? requestedLang : DEFAULT_LANG
+
+if (requestedLang) {
+  i18n.changeLanguage(lang)
+}
+
 if (redirectUrl) {
-  const selfCareIdentityToken = window.location.hash.replace('#id=', '')
   const url = `/ui/${i18n.language}${redirectUrl}#id=${selfCareIdentityToken}`
```
Alepazz: `i18n.changeLanguage(lang)` returns a Promise — i18next typically updates `language` synchronously, but relying on that is fragile. You already have the validated `lang` variable in scope; using it here decouples the URL from the timing of `changeLanguage`.

```suggestion
  const url = `/ui/${lang}${redirectUrl}#id=${selfCareIdentityToken}`
```

src/App.tsx
```diff
@@ -17,15 +17,31 @@ import { queryClient } from './config/query-client'
...
+
+  window.location.replace(url)
+} else if (requestedLang) {
+  fragmentParams.delete('lang')
+  const url = `/ui/${i18n.language}/catalogo-e-service#${fragmentParams.toString()}`
```
Alepazz: Same as above — prefer `lang` over `i18n.language` for the redirect URL.

```suggestion
  const url = `/ui/${lang}/catalogo-e-service#${fragmentParams.toString()}`
```

src/App.tsx
```diff
@@ -17,15 +17,31 @@ import { queryClient } from './config/query-client'
...
 import i18n from './config/react-i18next'
+import { DEFAULT_LANG, LANGUAGES } from './config/constants'
 
 // --- Init application ----
 
 const urlParams = new URLSearchParams(window.location.search)
 const redirectUrl = urlParams.get('redirectUrl')
 
+const fragmentParams = new URLSearchParams(window.location.hash.replace('#', ''))
+const selfCareIdentityToken = fragmentParams.get('id') ?? ''
```
Alepazz: `selfCareIdentityToken` is only consumed inside the `if (redirectUrl)` branch — in the `else if (requestedLang)` branch the `id` fragment is preserved transparently via `fragmentParams.toString()`. Consider scoping the constant inside the redirect branch where it's actually used.

src/api/auth/__tests__/auth.services.test.ts
```diff
@@ -0,0 +1,160 @@
...
+      const getSessionToken = await importGetSessionToken()
+      const result = await getSessionToken()
+
+      expect(result).toBe(mockSessionToken)
+      expect(window.localStorage.getItem(STORAGE_KEY_SESSION_TOKEN)).toBe(mockSessionToken)
+    })
+  })
+
+  describe('fragment parsing without lang parameter', () => {
+    it('should only  the id token when hash contains only #id=<token>', async () => {
```
Alepazz: Typo: missing verb (likely "extract") and double space.

```suggestion
    it('should only extract the id token when hash contains only #id=<token>', async () => {
```

src/api/auth/auth.services.ts
```diff
@@ -46,10 +46,11 @@ async function getSessionToken(): Promise<string | null> {
 
   // 3. See if we are trying to login as support operator
   // If the url has contains saml2 and jwt, we are trying to login as support operator
```
Alepazz: Pre-existing typo, easy to fix in this refactor: "has contains" → "contains".

```suggestion
  // If the url contains saml2 and jwt, we are trying to login as support operator
```

Alepazz (approved):
LGTM. Left some inline comments to verify.

PR #1802, opened 2026-04-02
chore: removed FEATURE_FLAG_NOTIFICATION_CONFIG (PIN-9636)

l-giacomini:
## 🔗 Issue
[PIN-9636](https://pagopa.atlassian.net/browse/PIN-9636)

## 📝 Description / Context
FEATURE_FLAG_NOTIFICATION_CONFIG needs to be removed, its value will always be `true`

## 🛠 List of changes
- Removed FEATURE_FLAG_NOTIFICATION_CONFIG
- Kept logic based on flag's `true` value

[PIN-9636]: https://pagopa.atlassian.net/browse/PIN-9636?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/router/components/RoutesWrapper/AuthGuard.tsx
```diff
@@ -76,7 +75,7 @@ export const AuthGuard: React.FC<AuthGuardProps> = ({
     const notificationsRoute: Array<RouteKey> = ['NOTIFICATIONS', 'NOTIFICATIONS_CONFIG']
 
     // return
-    return !notificationsRoute.includes(routeKey) || FEATURE_FLAG_NOTIFICATION_CONFIG
+    return !notificationsRoute.includes(routeKey)
```
Alepazz: The previous expression `!notificationsRoute.includes(routeKey) || FEATURE_FLAG_NOTIFICATION_CONFIG` always evaluated to `true` (since the flag was always `true`). Simplifying it to `!notificationsRoute.includes(routeKey)` changes the behavior: notification routes now return `false`, which causes `ForbiddenError` for all authenticated users trying to access NOTIFICATIONS or NOTIFICATIONS_CONFIG.

Since the flag was always `true` and the function always returned `true`, the fix should be to remove `isUserAllowedToAccessNotificationPage` entirely and drop its check from the guard condition (line 88). What do you think?

l-giacomini: You're right, I've removed the isUserAllowedToAccessNotificationPage method and its check in: https://github.com/pagopa/pdnd-interop-frontend/pull/1802/changes/d721974533751e8d5c00596cc94cdc4fccdc2cad

Alepazz (approved):
LGTM!

src/pages/PartyRegistryPage/components/PartyContactsSection/PartyContactsSection.tsx
```diff
@@ -18,7 +17,7 @@ export const PartyContactsSection: React.FC = () => {
 
   const { isAdmin } = AuthHooks.useJwt()
 
-  const isUserEnabledToShowNotificationConfig = FEATURE_FLAG_NOTIFICATION_CONFIG && isAdmin
+  const isUserEnabledToShowNotificationConfig = isAdmin
```
borgesis95: this var (`isUserEnabledToShowNotificationConfig`)  is not necessary anymore, I think that we can use isAdmin inline . Can you remove it?

src/router/components/RoutesWrapper/AuthGuard.tsx
```diff
@@ -72,20 +71,12 @@ export const AuthGuard: React.FC<AuthGuardProps> = ({
     return isOrganizationAllowedToDelegations || !delegationsRoutes.includes(routeKey)
   }
 
-  function isUserAllowedToAccessNotificationPage() {
```
borgesis95, line 75: why here we're going to remove the function?

borgesis95 (approved):

src/router/components/RoutesWrapper/AuthGuard.tsx
l-giacomini: The deletion of this function comes from a comment from @Alepazz (https://github.com/pagopa/pdnd-interop-frontend/pull/1802#discussion_r3028299069):

> The previous expression !notificationsRoute.includes(routeKey) || FEATURE_FLAG_NOTIFICATION_CONFIG always evaluated to true (since the flag was always true). Simplifying it to !notificationsRoute.includes(routeKey) changes the behavior: notification routes now return false, which causes ForbiddenError for all authenticated users trying to access NOTIFICATIONS or NOTIFICATIONS_CONFIG.

> Since the flag was always true and the function always returned true, the fix should be to remove isUserAllowedToAccessNotificationPage entirely and drop its check from the guard condition (line 88).

src/pages/PartyRegistryPage/components/PartyContactsSection/PartyContactsSection.tsx
l-giacomini: Fixed in: https://github.com/pagopa/pdnd-interop-frontend/pull/1802/changes/a1b7a0290b869ccf08132f626e8800f69539560f

borgesis95 (approved):

PR #1811, opened 2026-04-07
fix: update IconLink components to use button (PIN-2603 - A11Y)

martinaCampoli:
## 🔗 Issue

[PIN-A2-2603](https://pagopa.atlassian.net/browse/A2-2603)

## 📝 Description / Context

This pull request makes a small update to the `ConsumerAgreementDetailsGeneralInfoSection` component to improve accessibility and semantic correctness. The `IconLink` components are now explicitly rendered as `<button>` elements instead of their previous default.

BEFORE:

https://github.com/user-attachments/assets/89dcc4fb-8857-4c5e-a6a1-388897b03c4f

NOW:

https://github.com/user-attachments/assets/da273cfd-5c33-4ef0-b404-d66e4a792762

### DIFF
--- a/src/pages/ConsumerAgreementDetailsPage/components/ConsumerAgreementDetailsGeneralInfoSection/ConsumerAgreementDetailsGeneralInfoSection.tsx
+++ b/src/pages/ConsumerAgreementDetailsPage/components/ConsumerAgreementDetailsGeneralInfoSection/ConsumerAgreementDetailsGeneralInfoSection.tsx
@@ -134,7 +134,6 @@ export const ConsumerAgreementDetailsGeneralInfoSection: React.FC = () => {
                   component="button"
                   sx={{ alignSelf: 'start' }}
                   onClick={handleOpenCertifiedAttributesDrawer}
-                  style={{ alignSelf: 'start' }}
                 >
                   {t('certifiedAttributeLink.label')}
                 </IconLink>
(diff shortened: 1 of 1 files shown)


src/pages/ConsumerAgreementDetailsPage/components/ConsumerAgreementDetailsGeneralInfoSection/ConsumerAgreementDetailsGeneralInfoSection.tsx
```diff
@@ -131,6 +131,7 @@ export const ConsumerAgreementDetailsGeneralInfoSection: React.FC = () => {
               <>
                 <IconLink
                   startIcon={<RuleIcon />}
+                  component="button"
                   sx={{ alignSelf: 'start' }}
                   onClick={handleOpenCertifiedAttributesDrawer}
```
Alepazz, line 136: **Tip:** I agree with Copilot — `alignSelf: 'start'` is duplicated between `sx` and `style` here. Since you're already touching these lines, it might be worth cleaning this up: remove `style` and keep only `sx` on this IconLink, and convert the `style` on the other IconLink (L146) to `sx` for consistency. Slightly out of scope but a quick win.

Alepazz (approved):
LGTM. I agree with Copilot's observation about the duplicated `alignSelf` styling — see tip below.

borgesis95 (approved):

PR #1846, opened 2026-04-21
feature/PIN-2666_INT-97-tab-navigation-tooltip-in-update-documetation-drawer

martinaCampoli:
## 🔗 Issue

[PIN-A2-2666](https://pagopa.atlassian.net/browse/A2-2666)

## 📝 Description / Context

**Accessibility improvements:**
* Updated the `InfoIcon` in `InterfaceDocumentContainer` to be focusable, with `tabIndex={0}`, `role="button"`, `aria-disabled={true}`, and an appropriate `aria-label` for screen readers. This makes the icon accessible while communicating that it is not interactive. 
Screenshot 2026-04-21 alle 15 26 05


Alepazz (approved):
LGTM — nice work applying Copilot's suggestion for the wrapper approach, clean and well-scoped changes.

borgesis95 (approved):

PR #1848, opened 2026-04-22
feature/PIN-2671_INT-102-tab-navigation-add-operators-to-client-alert

martinaCampoli:
## 🔗 Issue

[PIN-A2-2671](https://pagopa.atlassian.net/browse/A2-2671)

## 📝 Description / Context

Added aria-live attribute to the admin alert in AddOperatorsToClientDrawer and in AddUsersToKeychainDrawer for accessibility reasons.

Now the screen reader announces the alert content to the user:
Screenshot 2026-04-22 alle 14 58 04

### DIFF
--- a/src/components/shared/AddOperatorsToClientDrawer.tsx
+++ b/src/components/shared/AddOperatorsToClientDrawer.tsx
@@ -88,7 +88,7 @@ export const AddOperatorsToClientDrawer: React.FC<AddOperatorsDrawerProps> = ({
             loading={isLoadingAllPartyOperators}
           />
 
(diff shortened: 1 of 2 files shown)


Alepazz (approved):
Copilot's comment about the `role="alert"` + `aria-live="polite"` conflict is not fully resolved — MUI's `<Alert>` renders `role="alert"` by default, so removing the explicit attribute doesn't change the rendered output. Adding `role="status"` would override the default and fix the mismatch.

borgesis95 (approved):

PR #1852, opened 2026-04-24
feature/PIN-2683_INT-47-missing-attribute

martinaCampoli:
## 🔗 Issue

[PIN-A2-2683](https://pagopa.atlassian.net/browse/A2-2683)

## 📝 Description / Context
This pull request introduces accessibility improvements to form label components by adjusting their rendered HTML elements to ensure proper semantics and attributes. 

* In `RiskAnalysisInputWrapper.tsx`, the `FormLabel` now always uses the `legend` component for better accessibility.
* In `RHFRadioGroup.tsx`, the `FormLabel` is set to render as a `div` to avoid rendering a `label` without a `for` attribute.


Alepazz (commented):
The approach addresses a valid a11y concern, but the current fixes introduce new semantic issues. See Copilot's comments — I agree with all three.

borgesis95 (approved):

PR #1859, opened 2026-04-28
bugfix:  fe reports resolution (PIN-9731)

antofo:
## Jira Issue
[PIN-9731](https://pagopa.atlassian.net/browse/PIN-9731)

## Context/Why
**Eservice/Template EService**
- Minor UI fixes

## Key Changes

[PIN-9731]: https://pagopa.atlassian.net/browse/PIN-9731?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/pages/ProviderEServiceTemplateSummaryPage/ProviderEServiceTemplateSummary.page.tsx
+++ b/src/pages/ProviderEServiceTemplateSummaryPage/ProviderEServiceTemplateSummary.page.tsx
@@ -203,7 +203,7 @@ const ProviderEServiceTemplateSummaryPage: React.FC = () => {
             </Stack>
           </Alert>
         )}
-        {hasMissingFields && !isLoading && (
+        {!canBePublished() && !isLoading && (
           <Alert severity="warning" sx={{ mt: 3 }}>
             {t('summary.missingFieldsBanner')}
           </Alert>
--- a/src/pages/ProviderEServiceTemplateSummaryPage/components/ProviderEServiceTemplateThresholdsAndAttributesSummarySection.tsx
+++ b/src/pages/ProviderEServiceTemplateSummaryPage/components/ProviderEServiceTemplateThresholdsAndAttributesSummarySection.tsx
@@ -167,7 +167,7 @@ const TemplateSummaryAttributeGroup: React.FC<TemplateSummaryAttributeGroupProps
             </Box>
             {index < attributes.length - 1 && (
               <Divider sx={{ py: 1 }}>
-                <Typography color="text.primary" fontWeight={700} textTransform={'uppercase'}>
+                <Typography color="text.primary" fontWeight={700} textTransform="uppercase">
                   {tAttribute('or')}
                 </Typography>
               </Divider>
--- a/src/pages/ProviderEServiceTemplateSummaryPage/components/ProviderEServiceTemplateThresholdsAndAttributesSummarySection.tsx
+++ b/src/pages/ProviderEServiceTemplateSummaryPage/components/ProviderEServiceTemplateThresholdsAndAttributesSummarySection.tsx
@@ -144,6 +144,8 @@ const TemplateSummaryAttributeGroup: React.FC<TemplateSummaryAttributeGroupProps
     keyPrefix: 'summary.thresholdsAndAttributesSummary.attributes',
   })
 
+  const { t: tAttribute } = useTranslation('attribute')
+
   return (
     <AttributeGroupContainer
       title={t('requirementTitle', {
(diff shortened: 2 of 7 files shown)


src/pages/ProviderEServiceTemplateSummaryPage/ProviderEServiceTemplateSummary.page.tsx
```diff
@@ -203,7 +203,7 @@ const ProviderEServiceTemplateSummaryPage: React.FC = () => {
             </Stack>
           </Alert>
         )}
-        {hasMissingFields && !isLoading && (
+        {!canBePublished() && !isLoading && (
```
Alepazz, line 206: The condition switched from `hasMissingFields` to `!canBePublished()`, which is also true when the interface upload is missing or when personal data is not set. A few questions:

- When `arePersonalDataSet` is false (with the feature flag enabled), the dedicated alert above and this banner now show together. Is that intentional?
- When only the interface is missing, the copy "fill in the missing information" can be misleading since the user might look for empty form fields rather than a missing upload. Should the banner copy be revisited, or split into separate alerts per cause?

src/pages/ProviderEServiceTemplateSummaryPage/components/ProviderEServiceTemplateThresholdsAndAttributesSummarySection.tsx
```diff
@@ -158,10 +160,19 @@ const TemplateSummaryAttributeGroup: React.FC<TemplateSummaryAttributeGroupProps
...
-        {attributes.map((attribute) => (
-          <Box key={attribute.id} component="li">
-            <AttributeContainer attribute={attribute} />
-          </Box>
+        {attributes.map((attribute, index) => (
+          <React.Fragment key={attribute.id}>
+            <Box key={attribute.id} component="li">
+              <AttributeContainer attribute={attribute} />
+            </Box>
+            {attributes.length > 1 && index < attributes.length - 1 && (
```
Alepazz: The `attributes.length > 1` check is redundant: when `attributes.length === 1` the only iteration has `index === 0` and `0 < 0` is already false.

```suggestion
            {index < attributes.length - 1 && (
```

src/pages/ProviderEServiceTemplateSummaryPage/components/ProviderEServiceTemplateThresholdsAndAttributesSummarySection.tsx
```diff
@@ -158,10 +160,19 @@ const TemplateSummaryAttributeGroup: React.FC<TemplateSummaryAttributeGroupProps
...
+              <Divider sx={{ py: 1 }}>
+                <Typography color="text.primary" fontWeight={700} textTransform={'uppercase'}>
```
Alepazz: Nit: string literal doesn't need braces.

```suggestion
                <Typography color="text.primary" fontWeight={700} textTransform="uppercase">
```

src/static/locales/it/attribute.json
```diff
@@ -42,7 +42,7 @@
     "manage": {
       "success": {
         "provider": "Il fruitore possiede almeno uno dei seguenti attributi",
-        "consumer": "Possiedi almeno uno di questi attributi."
+        "consumer": "Possiedi almeno uno dei seguenti attributi"
```
Alepazz, line 45: The IT copy was polished here (no trailing period, "questi" → "dei seguenti"). The EN counterparts at `en/attribute.json:45` ("...one of these attributes.") and `:63` ("...any of these attributes.") still have the trailing period and the "these" wording. Was the polish intentionally IT-only, or should EN be aligned?

Alepazz (commented):
Left a few observations/questions inline

src/pages/ProviderEServiceTemplateSummaryPage/components/ProviderEServiceTemplateThresholdsAndAttributesSummarySection.tsx
antofo: You're right I changed it.

src/static/locales/it/attribute.json
antofo: I aligned the copy in the EN version because it makes sense. I think there are no official guidelines other then simply translate the IT version. In this case it was reported from QA only the wrong IT copy.

src/pages/ProviderEServiceTemplateSummaryPage/components/ProviderEServiceTemplateThresholdsAndAttributesSummarySection.tsx
antofo: resolved

src/pages/ProviderEServiceTemplateSummaryPage/ProviderEServiceTemplateSummary.page.tsx
borgesis95: I'll try to address both (good) points:

- For me is good that user (in that case) see both alerts.
-  For this case i think that message is right. Remember that each section has chip which describe in detail what field is missing. @Alepazz

Alepazz: Ok seems good to me! Thanks

borgesis95 (approved):

Alepazz (approved):
LGTM. Two minor leftovers worth a quick follow-up:

- `en/attribute.json` (`success.consumer`): the EN alignment was applied to `error.consumer` but missed here — still reads `"You possess at least one of these attributes."` with trailing period and `these` instead of `the following`.
- `it/eserviceTemplate.json`: `"Attributi Verificati"` and `"Attributi Dichiarati"` are now capitalized while the sibling `certified.label` stays lowercase (`"Attributi certificati"`). The current mix looks unintentional.

PR #1867, opened 2026-04-30
Feature/pin 9937 page container redesign

alten-dturus:
## 🔗 Issue

[PIN-9937](https://pagopa.atlassian.net/browse/PIN-9937)

## 📝 Description / Context

Implement component NewPageContainer that is a redesign of the PageContaier to include the new secondary header for eservice version. The new component can be used also for eservice information in the agreement details page or purpose details page but for now it is implemented only in consumer and provider eservice details pages. For now it is used only in eservice details pages

## 🛠 List of changes

- added new strings for version label
- implement NewPageContainer component to reflect the design.
- implemented new PageContainerSecondaryIntro component that show data about version or eventually eservice with some personalization allowed

## 🧪 How to test

- Open an e-service detail page.
- The component should show the eservice name, state and actions on the first row and a box with version informations and eventually actions
- Verify Italian and English copy.

## ⚠️ Notes

The get actions hooks are not updated because the BE is not ready so the integration is not done yet

[PIN-9937]: https://pagopa.atlassian.net/browse/PIN-9937?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/layout/containers/NewPageContainer.tsx
+++ b/src/components/layout/containers/NewPageContainer.tsx
@@ -60,7 +60,7 @@ type IntroProps = {
   infoSection?: HeaderInfoSectionProps
 } & ActionsSectionProps
 
-type PageContainerProps = {
+export type PageContainerProps = {
   isLoading?: boolean
   sx?: SxProps
   children: React.ReactNode
--- a/src/components/layout/containers/NewPageContainer.tsx
+++ b/src/components/layout/containers/NewPageContainer.tsx
@@ -219,7 +219,7 @@ const ActionsSection: React.FC<ActionsSectionProps> = ({
       <Stack direction="row" spacing={2}>
         {secondaryAction && renderActionButton(secondaryAction)}
         {primaryAction && renderActionButton(primaryAction)}
-        {menuActions.length !== 0 && <ActionMenu actions={menuActions} />}
+        {menuActions.length !== 0 && <ActionMenu actions={menuActions} hasIconBorder={true} />}
       </Stack>
     </Stack>
   )
--- a/src/components/layout/containers/NewPageContainer.tsx
+++ b/src/components/layout/containers/NewPageContainer.tsx
@@ -1,0 +1,273 @@
+import React from 'react'
+import type { SxProps } from '@mui/material'
+import { Box, Button, Skeleton, Stack, Tooltip, Typography } from '@mui/material'
+import type { ActionItemButton } from '@/types/common.types'
+import { Breadcrumbs } from '../Breadcrumbs'
+import { StatusChip } from '@/components/shared/StatusChip'
+import { Link, type RouteKey } from '@/router'
+import ArrowBackIcon from '@mui/icons-material/ArrowBack'
+import { ActionMenu, ActionMenuSkeleton } from '@/components/shared/ActionMenu'
+
+export type PageBackToAction = {
+  label: string
+  to: RouteKey
+  params?: Record<string, string>
+  urlParams?: Record<string, string>
+}
+
+type PageContainerActionsProps = {
+  topSideActions?: Array<ActionItemButton>
+}
+
+type PageContainerBreadcrumbsProps = {
+  backToAction?: PageBackToAction
+}
+
+type PageContainerSecondaryIntroProps = {
+  label: string
+  link: {
+    label: string
+    onClink: () => void
+  }
+  actions: Array<ActionItemButton>
+  statusChip?: React.ComponentProps<typeof StatusChip>
+}
+
+type PageContainerIntroProps = {
+  title?: string
+  description?: string | React.ReactNode
+  secondaryIntro?: PageContainerSecondaryIntroProps
+} & PageContainerActionsProps
+
+type PageContainerProps = {
+  isLoading?: boolean
+  sx?: SxProps
+  children: React.ReactNode
+} & PageContainerBreadcrumbsProps &
+  PageContainerIntroProps
+
+type PageContainerSkeletonProps = {
+  children?: React.ReactNode
+  backToAction?: PageBackToAction
+}
+
+export const NewPageContainer: React.FC<PageContainerProps> = ({
+  children,
+  isLoading,
+  ...props
+}) => {
+  return (
+    <Stack direction={'column'} spacing={3}>
+      <PageContainerBreadcrumbs {...props} />
+      {isLoading ? <PageContainerIntroSkeleton /> : <PageContainerIntro {...props} />}
+      {/* {!isLoading && <PageContainerActions {...props} />} */}
+      <Box>{children}</Box>
+    </Stack>
+  )
+}
+
+export const PageContainerSkeleton: React.FC<PageContainerSkeletonProps> = ({
+  children,
+  backToAction,
+}) => {
+  return (
+    <Box>
+      <PageContainerBreadcrumbs backToAction={backToAction} />
+      <PageContainerIntroSkeleton />
+      <Box sx={{ mt: 1 }}>{children}</Box>
+    </Box>
+  )
+}
+
+const PageContainerIntro: React.FC<PageContainerIntroProps> = ({
+  title,
+  description,
+  topSideActions,
+  secondaryIntro,
+}) => {
+  return (
+    <Box sx={{ flex: 1 }}>
+      <Stack direction="row" spacing={2}>
+        {title && (
+          <Typography component="h1" variant="h4">
+            {title}
+          </Typography>
+        )}
+        {<PageContainerActions topSideActions={topSideActions} />}
+      </Stack>
+      {description && <PageContainerSubtitle description={description} />}
+      {secondaryIntro && <PageContainerSecondaryIntro {...secondaryIntro} />}
+    </Box>
+  )
+}
+
+type PageContainerSubtitle = {
+  description: string | React.ReactNode
+}
+
+const PageContainerSubtitle: React.FC<PageContainerSubtitle> = ({ description }) => {
+  return typeof description === 'string' ? (
+    <Typography component="p" variant="body1" sx={{ mt: 1, mb: 0 }}>
+      {description}
+    </Typography>
+  ) : (
+    description
+  )
+}
+const PageContainerBreadcrumbs: React.FC<PageContainerBreadcrumbsProps> = ({ backToAction }) => {
+  return (
+    <Stack direction="row" alignItems="center" spacing={2} sx={{ mb: 1 }}>
+      {backToAction && (
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
+        <Link
+          to={backToAction.to}
+          // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+          // @ts-ignore
+          params={backToAction.params}
+          options={backToAction.urlParams ? { urlParams: backToAction.urlParams } : undefined}
+          as="button"
+          startIcon={<ArrowBackIcon />}
+          size="small"
+          variant="naked"
+        >
+          {backToAction.label}
+        </Link>
+      )}
+      <Breadcrumbs />
+    </Stack>
+  )
+}
+
+const PageContainerActions: React.FC<PageContainerActionsProps> = ({ topSideActions }) => {
+  if (!topSideActions) return null
+
+  const primaryActions = topSideActions?.filter(
+    (action) => action.hierarchy && action.hierarchy === 'primary'
+  )
+  const secondaryActions = topSideActions?.filter(
+    (action) => action.hierarchy && action.hierarchy === 'secondary'
+  )
+  const menuActions = topSideActions?.filter((action) => action.hierarchy === undefined)
+
+  const getButtonWrapper = (tooltip?: string, disabled?: boolean) => {
+    return tooltip
+      ? ({ children }: { children: React.ReactElement }) => (
+          <Tooltip arrow title={tooltip}>
+            <span tabIndex={disabled ? 0 : undefined}>{children}</span>
+          </Tooltip>
+        )
+      : React.Fragment
+  }
+
+  return (
+    <Stack
+      sx={{ minHeight: 40 }}
+      direction="row"
+      alignItems="center"
+      justifyContent="flex-end"
+      flex={1}
+    >
+      <Stack direction="row" spacing={2}>
+        {secondaryActions &&
+          secondaryActions.map(({ action, label, color, icon: Icon, tooltip, ...props }, index) => {
+            const Wrapper = getButtonWrapper(tooltip, props.disabled)
+            return (
+              <Wrapper key={index}>
+                <Button
+                  onClick={action}
+                  variant="outlined"
+                  size="small"
+                  color={color}
+                  startIcon={Icon && <Icon />}
+                  {...props}
+                >
+                  {label}
+                </Button>
+              </Wrapper>
+            )
+          })}
+        {primaryActions &&
+          primaryActions.map(({ action, label, color, icon: Icon, tooltip, ...props }, index) => {
+            const Wrapper = getButtonWrapper(tooltip, props.disabled)
+            return (
+              <Wrapper key={index}>
+                <Button
+                  onClick={action}
+                  variant="contained"
+                  size="small"
+                  color={color}
+                  startIcon={Icon && <Icon />}
+                  {...props}
+                >
+                  {label}
+                </Button>
+              </Wrapper>
+            )
+          })}
+        {menuActions ? <ActionMenu actions={menuActions} /> : <ActionMenuSkeleton />}
+      </Stack>
+    </Stack>
+  )
+}
+
+export const PageContainerSecondaryIntro: React.FC<PageContainerSecondaryIntroProps> = ({
+  label,
+  link,
+  statusChip,
+  actions,
+}) => {
+  return (
+    <Box bgcolor="ThreeDFace" py={2} px={2} borderRadius={1} mt={3}>
+      <Stack direction="row" alignItems="center" justifyContent="space-between">
+        <Stack direction="row" spacing={4} alignItems="center">
+          <Typography component="h2" variant="body2" textTransform="uppercase">
+            {label}
+          </Typography>
+          <Button
+            component="a"
+            type="button"
+            variant="naked"
+            sx={{ textDecoration: 'underline' }}
+            onClick={link.onClink}
+          >
+            {link.label}
+          </Button>
+          {statusChip && <StatusChip {...statusChip} />}
+        </Stack>
+        <Stack direction="row" spacing={3}>
+          {actions.map(({ action, label, color, icon: Icon, ...props }, index) => {
+            return (
+              <Button
+                key={index}
+                onClick={action}
+                variant="text"
+                size="small"
+                color={color}
+                startIcon={Icon && <Icon />}
+                {...props}
+              >
+                {label}
+              </Button>
+            )
+          })}
+        </Stack>
+      </Stack>
+    </Box>
+  )
+}
+
+export const PageContainerIntroSkeleton: React.FC = () => {
+  return (
+    <Stack direction="row" alignItems="end" spacing={2}>
+      <Box sx={{ flex: 1 }}>
+        <Typography component="h1" variant="h4">
+          <Skeleton />
+        </Typography>
+        <Typography component="p" variant="body1" sx={{ mt: 1, mb: 0 }}>
+          <Skeleton />
+        </Typography>
+      </Box>
+    </Stack>
+  )
+}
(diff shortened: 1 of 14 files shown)


src/components/layout/containers/NewPageContainer.tsx
```diff
@@ -0,0 +1,273 @@
...
+
+type PageContainerBreadcrumbsProps = {
+  backToAction?: PageBackToAction
+}
+
+type PageContainerSecondaryIntroProps = {
+  label: string
+  link: {
+    label: string
+    onClink: () => void
```
Alepazz: Typo: `onClink` → `onClick`.

```suggestion
    onClick: () => void
```

src/components/layout/containers/NewPageContainer.tsx
```diff
@@ -0,0 +1,273 @@
+import React from 'react'
```
Alepazz, line 1: As an alternative to forking the file: would extending `PageContainer` in-place with an optional `secondaryIntro` prop have worked? Existing call sites that don't pass it would render unchanged, while the e-service detail pages would opt in to the new box. Same end result without two files to keep in sync.

src/components/layout/containers/NewPageContainer.tsx
```diff
@@ -0,0 +1,273 @@
...
+    <Stack direction={'column'} spacing={3}>
+      <PageContainerBreadcrumbs {...props} />
+      {isLoading ? <PageContainerIntroSkeleton /> : <PageContainerIntro {...props} />}
+      {/* {!isLoading && <PageContainerActions {...props} />} */}
+      <Box>{children}</Box>
+    </Stack>
+  )
+}
+
+export const PageContainerSkeleton: React.FC<PageContainerSkeletonProps> = ({
```
Alepazz, line 93: Heads-up: `PageContainerSkeleton` and `PageContainerIntroSkeleton` are now exported with the same names from both `PageContainer.tsx` and this file. `containers/index.ts` only re-exports from `PageContainer`, so today there's no breakage — but if anyone later adds `export * from './NewPageContainer'` they'll hit a silent re-export collision.

src/components/layout/containers/NewPageContainer.tsx
```diff
@@ -0,0 +1,273 @@
...
+}
+
+export const PageContainerSecondaryIntro: React.FC<PageContainerSecondaryIntroProps> = ({
+  label,
+  link,
+  statusChip,
+  actions,
+}) => {
+  return (
+    <Box bgcolor="ThreeDFace" py={2} px={2} borderRadius={1} mt={3}>
```
Alepazz: `ThreeDFace` is a deprecated CSS system color: it renders inconsistently across browsers and doesn't respond to the MUI theme. Consider replacing it with a theme token like `bgcolor: 'background.default'` or a dedicated PagoPA color.

src/components/layout/containers/NewPageContainer.tsx
```diff
@@ -0,0 +1,273 @@
...
+export const NewPageContainer: React.FC<PageContainerProps> = ({
+  children,
+  isLoading,
+  ...props
+}) => {
+  return (
+    <Stack direction={'column'} spacing={3}>
+      <PageContainerBreadcrumbs {...props} />
+      {isLoading ? <PageContainerIntroSkeleton /> : <PageContainerIntro {...props} />}
+      {/* {!isLoading && <PageContainerActions {...props} />} */}
```
Alepazz: Is this a leftover from the refactor (the actions are now rendered inside `PageContainerIntro`)? If so, can it be removed?

src/components/layout/containers/NewPageContainer.tsx
```diff
@@ -0,0 +1,273 @@
...
+      <Breadcrumbs />
+    </Stack>
+  )
+}
+
+const PageContainerActions: React.FC<PageContainerActionsProps> = ({ topSideActions }) => {
+  if (!topSideActions) return null
+
+  const primaryActions = topSideActions?.filter(
+    (action) => action.hierarchy && action.hierarchy === 'primary'
```
Alepazz: The truthiness check is redundant: `action.hierarchy === 'primary'` already implies `action.hierarchy` is truthy. Same applies to the secondary filter on the next line.

```suggestion
    (action) => action.hierarchy === 'primary'
```

src/components/layout/containers/NewPageContainer.tsx
```diff
@@ -0,0 +1,273 @@
...
+  )
+}
+
+export const PageContainerSkeleton: React.FC<PageContainerSkeletonProps> = ({
+  children,
+  backToAction,
+}) => {
+  return (
+    <Box>
+      <PageContainerBreadcrumbs backToAction={backToAction} />
```
Alepazz: The braces around the JSX element are redundant — you're already inside JSX context.

```suggestion
        <PageContainerActions topSideActions={topSideActions} />
```

src/components/layout/containers/NewPageContainer.tsx
```diff
@@ -0,0 +1,273 @@
...
+  backToAction?: PageBackToAction
+}
+
+export const NewPageContainer: React.FC<PageContainerProps> = ({
+  children,
+  isLoading,
+  ...props
+}) => {
+  return (
+    <Stack direction={'column'} spacing={3}>
```
Alepazz: Minor: for consistency with the rest of the file, this can be a plain string attribute.

```suggestion
    <Stack direction="column" spacing={3}>
```

Alepazz (commented):
Looks like solid work — a few inline considerations. Also worth adding tests for the new conditional branches before more call sites depend on this component.

src/components/layout/containers/NewPageContainer.tsx
```diff
@@ -0,0 +1,273 @@
...
+  backToAction?: PageBackToAction
+}
+
+type PageContainerSecondaryIntroProps = {
+  label: string
+  link: {
+    label: string
+    onClink: () => void
+  }
+  actions: Array<ActionItemButton>
```
Alepazz: Consider making this optional with a default of `[]`. Both call sites in this PR currently pass `actions: []` as a placeholder.

```suggestion
  actions?: Array<ActionItemButton>
```

src/components/layout/containers/NewPageContainer.tsx
```diff
@@ -0,0 +1,273 @@
...
+  ) : (
+    description
+  )
+}
+const PageContainerBreadcrumbs: React.FC<PageContainerBreadcrumbsProps> = ({ backToAction }) => {
+  return (
+    <Stack direction="row" alignItems="center" spacing={2} sx={{ mb: 1 }}>
+      {backToAction && (
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
```
Alepazz: These `// @ts-ignore` were carried over from the previous file. Since this is a new file, it could be a good moment to type them properly — either via `RouteKey` generics or a typed wrapper.

src/types/common.types.ts
```diff
@@ -48,10 +48,15 @@ export type Provider = 'provider'
 export type Consumer = 'consumer'
 export type ProviderOrConsumer = Provider | Consumer
 
-export type ActionItem = { action: VoidFunction; label: string; fontColor?: string }
+export type ActionItem = {
+  action: VoidFunction
+  label: string
+  fontColor?: string
+  icon?: SvgIconComponent
+  hierarchy?: 'primary' | 'secondary'
```
Alepazz: Discussion point: `hierarchy?: 'primary' | 'secondary'` relies on `undefined` to mean "menu action", but that contract lives only in the rendering logic of `NewPageContainer`. Would it be worth making the menu case explicit, e.g. `hierarchy: 'primary' | 'secondary' | 'menu'`? It would make the type self-documenting and let the action filter become an exhaustive `ts-pattern` match instead of a split between truthy/falsy checks.

src/components/layout/containers/NewPageContainer.tsx
alten-dturus: I know that, I've done like this because I hope the integration in all pages will be as soon as possible so for now I wanted to reduce the changes necessary in renaming

src/components/layout/containers/NewPageContainer.tsx
alten-dturus: It would have produced more changes than necessary by duplicating it because the prop would have had to be used also for the management of the actions and then for the chip issue. And it would have been tested for regression errors by QA

src/components/layout/containers/NewPageContainer.tsx
alten-dturus: I didn't do that because every time we use this prop we will pass an array of objects. The array will be populated or maybe empty based on the hook used to get the actions array

src/components/layout/containers/NewPageContainer.tsx
alten-dturus: I didn't found the color in the suggetions so I choose the one who was similar. I will search better

src/types/common.types.ts
alten-dturus: This was a refuse from when I implement that to make also old pages work without changes in actions. Now I can this but the problem is that hierarchy is not defined in old actions so I have to let it optional until we don't have a hierarcy for every action

src/components/layout/containers/NewPageContainer.tsx
Alepazz: Optional with default `[]` doesn't change the hook-driven pattern — call sites that have a real array from a hook keep passing it exactly as before. The benefit is just that the current placeholder usages can simply omit the prop, and any future placeholder use stays clean. It's purely additive — no existing or future caller is forced to change.

src/types/common.types.ts
Alepazz: Fair point on the migration cost. One option that avoids touching `ActionItemButton`: introduce a dedicated type just for the new container, e.g. `NewActionItemButton` with hierarchy: `'primary' | 'secondary' | 'menu'` required. Old pages keep using `ActionItemButton` unchanged; the new container types its actions prop as `NewActionItemButton[]`, which forces an explicit hierarchy and lets the filter become an exhaustive `ts-pattern` match instead of relying on the **undefined === menu** convention. When the migration completes, the two types can be merged and the old one dropped.

src/components/layout/containers/NewPageContainer.tsx
alten-dturus: Make sense, I will add it

Alepazz:
I closed the two threads on forking vs. extending and on the duplicated `PageContainerSkeleton` / `PageContainerIntroSkeleton` names — your reasoning is fair on both. Just a non-blocking note on top: since both decisions rest on the full migration to the new container landing shortly, would you mind opening a follow-up ticket (or leaving a TODO referencing one)? Just to make sure the eventual removal of `PageContainer.tsx` and the rename don't slip through.

src/components/layout/containers/NewPageContainer.tsx
Alepazz: Remember to close this comment as follow up when you have the correct grey name

alten-dturus:
> I closed the two threads on forking vs. extending and on the duplicated `PageContainerSkeleton` / `PageContainerIntroSkeleton` names — your reasoning is fair on both. Just a non-blocking note on top: since both decisions rest on the full migration to the new container landing shortly, would you mind opening a follow-up ticket (or leaving a TODO referencing one)? Just to make sure the eventual removal of `PageContainer.tsx` and the rename don't slip through.

I think a task could be opened

src/components/layout/containers/NewPageContainer.tsx
```diff
@@ -0,0 +1,273 @@
...
+    onClink: () => void
+  }
+  actions: Array<ActionItemButton>
+  statusChip?: React.ComponentProps<typeof StatusChip>
+}
+
+type PageContainerIntroProps = {
+  title?: string
+  description?: string | React.ReactNode
+  secondaryIntro?: PageContainerSecondaryIntroProps
```
borgesis95: Could we replace this with another name var? 
```suggestion
  InfoBanner?: PageContainerSecondaryIntroProps
```
What do you think about `InfoBanner` or `DetailInfoBanner`?

src/components/layout/containers/NewPageContainer.tsx
borgesis95: If the different between primary and secondary button is only about `outlined` and `contained` I'would change name of var. e.g : 
- `outlinedActions`
- `containedActions` 

I would follow same pattern for `hierarchy` key
what do you think?

src/components/layout/containers/NewPageContainer.tsx
```diff
@@ -0,0 +1,273 @@
...
+  return (
+    <Stack
+      sx={{ minHeight: 40 }}
+      direction="row"
+      alignItems="center"
+      justifyContent="flex-end"
+      flex={1}
+    >
+      <Stack direction="row" spacing={2}>
+        {secondaryActions &&
```
borgesis95: is extremely necessary split `secondaryActions` and `primaryActions`? what i see here is same code with only different in variant `props`

src/pages/ConsumerEServiceDetailsPage/ConsumerEServiceDetails.page.tsx
```diff
@@ -61,15 +61,24 @@ const ConsumerEServiceDetailsPage: React.FC = () => {
...
       backToAction={{
         label: t('actions.backToCatalogLabel'),
         to: 'SUBSCRIBE_CATALOG_LIST',
       }}
+      secondaryIntro={
+        descriptor
+          ? {
+              label: t('versionHeaderLabel'),
+              link: { label: descriptor.version, onClink: () => {} }, // TODO navigation function
+              actions: [], // TODO actions for secondHeader
```
borgesis95: will be actions (and other TODO) resolved in different PRs?

borgesis95 (changes requested):
several comments, I think that for this component is necessary add test

src/components/layout/containers/NewPageContainer.tsx
alten-dturus: For me it is not a banner so I would not use that name, but now I start thinking

src/components/layout/containers/NewPageContainer.tsx
alten-dturus: I use these name because design use them to define the actions in [this ](https://www.figma.com/design/CpRV3kPvFEWLXGtJUgWeZW/Interop-%E2%80%94-Delivery-FE---QA?node-id=5426-18445&t=oPD5ZlVZxRyGmzj3-0).

src/components/layout/containers/NewPageContainer.tsx
alten-dturus: Yes because secondaryActions are placed before primary in order

src/pages/ConsumerEServiceDetailsPage/ConsumerEServiceDetails.page.tsx
alten-dturus: Yes, for some TODOs we need the dialogs from other PRs and for the actions we need the API BE to be integrated

src/components/layout/containers/NewPageContainer.tsx
borgesis95: for this is not sufficient put in order when you pass props?

src/components/layout/containers/NewPageContainer.tsx
borgesis95: In this case I would take a different approach. For us this is the main difference: outlined or contained. What we see in FIGMA will change when will be used properly label on it (e.g "sospendi" or "archivia" etc...).

src/components/layout/containers/NewPageContainer.tsx
borgesis95: What I suggest is an idea (free to suggest something different) :) SecondaryIntro seems not good for this case IMHO.

src/components/layout/containers/NewPageContainer.tsx
alten-dturus: For me is more important to think about hierarchy because differentiation is based on this prop, maybe in the next design the secondary actions will be naked and the primary outlined and we have to change all of the var names

alten-dturus: We can remove new hierarchy prop too and base our consideration only on variant prop in actionItemButtons if there is present we take it and if not we put the action in the menu, there are multiple possibilities we can choose

src/components/layout/containers/NewPageContainer.tsx
Alepazz: Can I suggest `VersionInfoSection`?

alten-dturus: I thought at it but in the future pages like agreements details and purpose details this section will contain the eservice details (like we see [here](https://www.figma.com/design/CpRV3kPvFEWLXGtJUgWeZW/Interop-%E2%80%94-Delivery-FE---QA?node-id=5971-20063&t=oPD5ZlVZxRyGmzj3-0)) so I was not sure about its correctness. But if we consider the details about eservice like effectively details about the version we are referring in those pages, maybe this name can be appropriate

src/pages/ConsumerEServiceDetailsPage/ConsumerEServiceDetails.page.tsx
```diff
@@ -53,23 +53,36 @@ const ConsumerEServiceDetailsPage: React.FC = () => {
...
     eserviceId: descriptor?.eservice.id,
     descriptorId: descriptor?.id,
   })
 
   return (
-    <PageContainer
+    <NewPageContainer
       title={descriptor?.eservice.name || ''}
-      topSideActions={actions}
+      statusChip={descriptor && { for: 'eservice', state: descriptor.state }}
```
Alepazz: `statusChip` is passed twice on this page: here at the top level (rendered in the main Intro) and inside `infoSection` below (rendered in `HeaderInfoSection`). The Provider page only passes it inside `infoSection`. Likely a duplicate to remove.

src/components/layout/containers/NewPageContainer.tsx
```diff
@@ -0,0 +1,300 @@
...
+      sx={{ minHeight: 40 }}
+      direction="row"
+      alignItems="center"
+      justifyContent="flex-end"
+      flex={1}
+    >
+      <Stack direction="row" spacing={2}>
+        {secondaryAction && renderActionButton(secondaryAction)}
+        {primaryAction && renderActionButton(primaryAction)}
+        {menuActions ? <ActionMenu actions={menuActions} /> : <ActionMenuSkeleton />}
```
Alepazz: This conditional always renders `<ActionMenu>`: `menuActions` is typed as `Array<ActionItemButton>` (optional) but the hooks always return an array (`[]` or populated), never `undefined`. The `<ActionMenuSkeleton />` fallback is unreachable. The loading state is already handled one level up by `isLoading ? <IntroSkeleton /> : <Intro {...props} />`, so the inline skeleton is dead code. Either remove the fallback, or change the type to make `menuActions` truly optional and rework the logic.

src/pages/ProviderEServiceDetailsPage/ProviderEServiceDetails.page.tsx
```diff
@@ -27,36 +27,41 @@ const ProviderEServiceDetailsPage: React.FC = () => {
   const isEserviceFromTemplate = Boolean(descriptor?.templateRef)
 
   descriptor?.delegation
```
Alepazz: Suggestion: this statement (`descriptor?.delegation`) sits on its own line with no assignment or return, so it is a no-op. Looks like a leftover and can be removed.

Alepazz (changes requested):
A few new points after the latest changes, see inline.

Alepazz (approved):

src/components/layout/containers/NewPageContainer.tsx
```diff
@@ -0,0 +1,300 @@
...
+  )
+}
+
+const ActionsSection: React.FC<ActionsSectionProps> = ({
+  primaryAction,
+  secondaryAction,
+  menuActions = [],
+}) => {
+  if (menuActions.length === 0 && !primaryAction) return null
+
```
borgesis95: It's possible that menuActions is empty, primaryActions are not defined but secondaryActions are available?

src/components/layout/containers/NewPageContainer.tsx
```diff
@@ -0,0 +1,300 @@
...
+    </Stack>
+  )
+}
+
+const ActionsSection: React.FC<ActionsSectionProps> = ({
+  primaryAction,
+  secondaryAction,
+  menuActions = [],
+}) => {
+  if (menuActions.length === 0 && !primaryAction) return null
```
borgesis95: It's possible that `menuActions` is empty, `primaryActions` are not defined but `secondaryActions` are available?
If answer is yes, secondary action will not rendered

borgesis95 (changes requested):
remember to check what happen on actionMenu within table

borgesis95 (approved):

PR #1872, opened 2026-05-04
fix: added check about descriptors for rulesetExpiration

borgesis95:
## 🔗 Issue
[PIN-9966](https://pagopa.atlassian.net/browse/PIN-9966)

## 📝 Description / Context

This PR fixes the ruleset expiration evaluation in Provider E-Service Summary.

Before this change, `isRulesetExpired` was evaluated only on `riskAnalysis.rulesetExpiration` which could incorrectly block publication even when the e-service already had published descriptors (new version flow).

Now isRulesetExpired is evaluated as true only when:

1. there are no other descriptors already published descriptors.length === 0
2. at least one riskAnalysis.rulesetExpiration is expired
 
This aligns the behavior with the expected business rule for first publication vs new version publication.

[PIN-9966]: https://pagopa.atlassian.net/browse/PIN-9966?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/pages/ProviderEServiceSummaryPage/ProviderEServiceSummary.page.tsx
+++ b/src/pages/ProviderEServiceSummaryPage/ProviderEServiceSummary.page.tsx
@@ -72,6 +72,7 @@ const ProviderEServiceSummaryPage: React.FC = () => {
 
   const isEServiceFromTemplate = descriptor?.templateRef
 
+  // In this case "draft" version is the actual version that user is creating, so if the e-service has only one draft it means that it has never been published before.
   const hasOnlyOneDraft = descriptor?.eservice.descriptors.length === 0
 
   const handleDeleteDraft = () => {
@@ -196,8 +197,8 @@ const ProviderEServiceSummaryPage: React.FC = () => {
 
   const isRulesetExpired =
     // check if the e-service had already been published. In this case we have to skip ruleset expiration check (https://pagopa.atlassian.net/browse/PIN-9966)
-    descriptor?.eservice.descriptors.length === 0 &&
-    eserviceRiskAnalyses &&
+    hasOnlyOneDraft
+  eserviceRiskAnalyses &&
     eserviceRiskAnalyses.some(
       (riskAnalysis) =>
         riskAnalysis.rulesetExpiration && new Date(riskAnalysis.rulesetExpiration) < new Date()
--- a/src/pages/ProviderEServiceSummaryPage/ProviderEServiceSummary.page.tsx
+++ b/src/pages/ProviderEServiceSummaryPage/ProviderEServiceSummary.page.tsx
@@ -195,6 +195,8 @@ const ProviderEServiceSummaryPage: React.FC = () => {
   const eserviceRiskAnalyses = descriptor?.eservice.riskAnalysis
 
   const isRulesetExpired =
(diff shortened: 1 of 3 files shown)


src/pages/ProviderEServiceSummaryPage/ProviderEServiceSummary.page.tsx
```diff
@@ -195,6 +195,8 @@ const ProviderEServiceSummaryPage: React.FC = () => {
   const eserviceRiskAnalyses = descriptor?.eservice.riskAnalysis
 
   const isRulesetExpired =
+    // check if the e-service had already been published. In this case we have to skip ruleset expiration check (https://pagopa.atlassian.net/browse/PIN-9966)
+    descriptor?.eservice.descriptors.length === 0 &&
```
Alepazz: Adding to Copilot's reuse suggestion: when reusing `hasOnlyOneDraft` here it's worth noting that the name doesn't quite fit the new context — it reads as "has only one draft", but the condition (`descriptors.length === 0`) actually means "no published descriptors yet / first publication". Consider a rename like `isFirstPublication` or `hasNoPublishedDescriptors` to keep the intent readable for both this fix and the existing usages.

src/pages/ProviderEServiceSummaryPage/__tests__/ProviderEServiceSummary.page.test.tsx
```diff
@@ -238,4 +238,67 @@ describe('ProviderEServiceSummaryPage', () => {
...
+
+      renderWithApplicationContext(<ProviderEServiceSummaryPage />, {
+        withReactQueryContext: true,
+        withRouterContext: true,
+      })
+
+      const publishButton = screen.getByRole('button', { name: 'publish' })
+      expect(publishButton).toBeDisabled()
+    })
+
```
Alepazz, line 273: Consistency nit: the first test sets `mock.eservice.descriptors = []` explicitly, but this one relies on the default mock factory having `descriptors.length > 0`. If that default ever changes, this test would silently pass for the wrong reason. Suggest setting `descriptors` explicitly in the `eservice` override too (next to `riskAnalysis`).

src/pages/ProviderEServiceSummaryPage/ProviderEServiceSummary.page.tsx
```diff
@@ -195,6 +195,8 @@ const ProviderEServiceSummaryPage: React.FC = () => {
   const eserviceRiskAnalyses = descriptor?.eservice.riskAnalysis
 
   const isRulesetExpired =
+    // check if the e-service had already been published. In this case we have to skip ruleset expiration check (https://pagopa.atlassian.net/browse/PIN-9966)
```
Alepazz, line 199: Was inlining the full Jira URL intentional?

Alepazz (commented):
The fix looks good and addresses the bug correctly. I agree with the two findings raised by Copilot below (misleading comment + duplication with the existing `hasOnlyOneDraft`). A few additional observations inline.

Alepazz (approved):

PR #1899, opened 2026-05-12
ci: add eslint checks (PIN-10023)

DenisLantero:
## 🔗 Issue

[PIN-10023](https://pagopa.atlassian.net/browse/PIN-10023)

## 📝 Description / Context

This PR restores ESLint checks in the frontend project before push and during CI, so lint issues are detected before changes are merged.

It also fixes the ESLint and Prettier issues currently present on `dev`, bringing the project to a clean lint state, and updates the Husky shared hooks to remove deprecated initialization lines that produced warnings during commit.

## 🛠 List of changes

- Added `lint` and `lint:fix` npm scripts.
- Added a CI linting step before build and tests.
- Added a shared `pre-push` hook that runs ESLint autofix and blocks the push if files are modified or lint errors remain.
- Fixed existing ESLint and Prettier issues in the codebase.
- Removed deprecated Husky initialization from `pre-commit` and `commit-msg` hooks.

## 🧪 How to test

- Run `pnpm run lint`.
- Look at the [failed CI step](https://github.com/pagopa/pdnd-interop-frontend/actions/runs/25724098970/job/75532286031) before the lint-fix commit 
- Create a local commit and verify Husky no longer prints the deprecated hook warning.

## 📘 Notes

The Husky initialization fix is a bit out of scope for this PR, but since we are modifying husky scripts in this PR, and since it's a really small change I decided to put it in this PR, if you want a separate PR I can do that, let me know.

For info, this is the warning that was present on every commit:

 ```
 husky - DEPRECATED

Please remove the following two lines from .husky/commit-msg:

#!/usr/bin/env sh
. "$(dirname -- "$0")/_/husky.sh"

They WILL FAIL in v10.0.0
```

[PIN-10023]: https://pagopa.atlassian.net/browse/PIN-10023?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### CODE REVIEW

File: .husky/pre-push [hunk 1]
```diff
@@ -0,0 +1,35 @@
...
+
+log() {
+  echo "[pre-push] $1"
+}
+
+log "Running ESLint autofix..."
+git diff --binary > "$baseline_diff"
+
+set +e
+pnpm run lint:fix > "$lint_output" 2>&1
```
Alepazz: Wouldn't it make more sense to scope this to the files modified in the branch only, rather than running ESLint across the whole codebase on every push? With `lint-staged` or a `git diff --name-only origin/dev...HEAD` filter, the hook would stay fast and a developer would not be blocked by lint issues in code they have not touched. What is the reasoning behind running it on everything?

.husky/pre-push
```diff
@@ -0,0 +1,35 @@
...
+lint_exit_code=$?
+set -e
+sed 's/^/[pre-push] /' "$lint_output"
+
+git diff --binary > "$after_lint_diff"
+
+if ! cmp -s "$baseline_diff" "$after_lint_diff"; then
```
Alepazz, line 104: Using `lint:fix` here means the hook rewrites files in the developer's working tree, and if anything changes the push is aborted with "review and commit the fixes". Two side-effects worth considering:

1. The files are modified even when the push is rejected, so the developer sees a dirty working tree after pushing without having touched anything; if they had in-progress unstaged changes, the autofix can blend into those.
2. The fix-then-recommit-then-repush flow adds a round-trip on every push that needs fixing.

An alternative would be to run plain `lint` (check only) in the hook: it fails fast with the list of issues, the developer runs `lint:fix` manually when they want, and the working tree is never touched by the hook. Same number of steps when something needs fixing, but more transparent and non-intrusive. Worth a discussion.

Alepazz (commented):
Looks good overall, a couple of points on the pre-push hook worth discussing. See inline.

.husky/pre-push
DenisLantero: Yes I agree, in fact, I think we might even discuss to use this same pattern in the CI as well.
I've done it this way because this is how it's done in the backend project, but I agree that for lint it makes more sense to only run it on changed files once the codebase is clean.

.husky/pre-push
Alepazz: No action needed

.husky/pre-push
Alepazz: Agreed. @borgesis95 does this approach (scoping the hook to changed files once the codebase is clean) work for you as well?

borgesis95: Uhm I would keep linting run scoped on pre-push, but about CI should run linting for entire codebase.

DenisLantero: I updated the pre-push hook, now it only lints **CHANGED FILES.** 

The logic is:

- if the branch already exists on the remote, it compares the remote branch tip with the local branch tip being pushed. So on repeated pushes it only lints the TS/TSX files changed since the last push, instead of linting the whole feature branch again.
- if the branch does not exist on the remote yet, it compares the local branch with its base branch/upstream fallback. So on the first push it lints all TS/TSX files changed by the branch.

In short: first push checks the branch diff from its base, while following pushes check only the new diff from the remote branch to the local branch.

I think this approach would also be safe in CI if we apply it to the PR diff.

In CI we should not use the “since last push” logic from the local pre-push hook. For PR validation, the right comparison is always: PR base branch vs PR head branch.

That means CI would lint all TS/TSX files changed by the PR. For release PRs this is still safe, because the PR diff is computed against the release target branch, so the changed-file list contains the files that the release PR is actually introducing or modifying.

This prevents new lint issues from entering through the PR, considering that after this PR is merged there shouldn't be any other lint issues in `dev`, meaning that the only lint issues would be caused by new changes.

@borgesis95 @Alepazz do you think there could be any downsides if we decide to only lint changed files in CI?

borgesis95: It works for me :)

Alepazz: Agreed, this works for me too.

Alepazz (approved):
Great job!

borgesis95 (approved):

PR #1904, opened 2026-05-13
feat: add async fields to eservice summary (PIN-10030)

DenisLantero:
## 🔗 Issue

[PIN-10030](https://pagopa.atlassian.net/browse/PIN-10030)

## 📝 Description / Context

This PR updates the provider e-service creation/edit summary to show asynchronous exchange information and prevent publication when mandatory async data is incomplete.

## 🛠 List of changes

- Added “Exchange type” to the general information summary card.
- Added async fields to the technical specifications summary card: callback interface, async exchange configuration, and associated producer keychains.
- Preserved missing-data warnings for incomplete async fields.
- Updated publication validation so async e-services require all mandatory async information before publishing.
- Added IT/EN translations for the new summary labels.
- Added focused tests for synchronous, complete asynchronous, and incomplete asynchronous summary scenarios.

## 🧪 How to test

- Open the provider e-service summary page for a synchronous draft e-service. Verify that the “General information” card shows the exchange type as synchronous and that the “Technical specifications” card does not show async-only fields.
- Open the provider e-service summary page for a complete asynchronous draft e-service. Verify that the “General information” card shows the exchange type as asynchronous. In the “Technical specifications” card, verify that callback interface, response time, resource availability time, confirmation, bulk response, max result set, and associated producer keychains are displayed.
- Open the provider e-service summary page for an incomplete asynchronous draft e-service. Verify that missing async values are shown with the existing missing-information warning and that publishing is disabled until the mandatory async data is complete.

[PIN-10030]: https://pagopa.atlassian.net/browse/PIN-10030?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


borgesis95 (approved):

Alepazz (commented):
A few observations before merging, see inline comments.

src/pages/ProviderEServiceSummaryPage/ProviderEServiceSummary.page.tsx
```diff
@@ -63,6 +64,15 @@ const ProviderEServiceSummaryPage: React.FC = () => {
     EServiceQueries.getDescriptorProvider(eserviceId, descriptorId)
   )
 
+  const { data: associatedKeychains } = useQuery({
```
Alepazz: This `useQuery` overlaps with the one added in `ProviderEServiceDocumentationSummarySection` (same endpoint, but `limit: 1` vs `limit: 50`). Since the cache keys differ, React Query fires two separate network calls for the same resource when an async draft is opened. Is there a reason to keep them separate, or could the page reuse the section's query (or vice versa) and derive the count from `results.length` / `pagination.totalCount`?

src/pages/ProviderEServiceSummaryPage/components/ProviderEServiceDocumentationSummarySection.tsx
```diff
@@ -22,6 +23,16 @@ export const ProviderEServiceDocumentationSummarySection: React.FC = () => {
 
   const downloadDocument = EServiceDownloads.useDownloadVersionDocument()
   const voucherLifespan = secondsToMinutes(descriptor.voucherLifespan)
+  const isAsyncExchange = descriptor.eservice.asyncExchange
+
+  const { data: associatedKeychains } = useQuery({
+    ...KeychainQueries.getKeychainsList({
+      eserviceId: descriptor.eservice.id,
+      offset: 0,
+      limit: 50,
```
Alepazz: Hardcoded `limit: 50` with no pagination. Is it realistic for a producer to have more than 50 keychains associated to the same e-service? If yes, the extras are silently dropped from the summary, so it would be worth either paging through them or surfacing a "+N more" indicator using `pagination.totalCount`.

src/pages/ProviderEServiceSummaryPage/ProviderEServiceSummary.page.tsx
```diff
@@ -204,6 +214,23 @@ const ProviderEServiceSummaryPage: React.FC = () => {
...
+  const areAsyncExchangeFieldsSet =
+    !descriptor?.eservice.asyncExchange ||
+    Boolean(
+      descriptor.asyncExchangeCallbackInterface &&
+      descriptor.asyncExchangeProperties &&
+      hasAsyncExchangeNumber(descriptor.asyncExchangeProperties.responseTime) &&
+      hasAsyncExchangeNumber(descriptor.asyncExchangeProperties.resourceAvailableTime) &&
+      typeof descriptor.asyncExchangeProperties.confirmation === 'boolean' &&
+      typeof descriptor.asyncExchangeProperties.bulk === 'boolean' &&
+      hasAsyncExchangeNumber(descriptor.asyncExchangeProperties.maxResultSet) &&
```
Alepazz: While the keychains query is in flight, `associatedKeychains` is `undefined`, so this expression resolves to `0 > 0`, `areAsyncExchangeFieldsSet` is `false`, and the Publish button renders disabled for a brief moment even on a fully complete async draft. Is the flicker intentional, or should we gate on the query `isLoading` / `isPending` so the button stays in a neutral state until the keychains are known?

src/pages/ProviderEServiceSummaryPage/ProviderEServiceSummary.page.tsx
```diff
@@ -204,6 +214,23 @@ const ProviderEServiceSummaryPage: React.FC = () => {
         riskAnalysis.rulesetExpiration && new Date(riskAnalysis.rulesetExpiration) < new Date()
     )
```
Alepazz, line 231: The helper just wraps `typeof value === 'number'`, which is already inlined for `confirmation` and `bulk` a few lines below. Inlining the number checks too would keep the validation block consistent in style.

src/pages/ProviderEServiceSummaryPage/ProviderEServiceSummary.page.tsx
```diff
@@ -204,6 +214,23 @@ const ProviderEServiceSummaryPage: React.FC = () => {
...
+    return typeof value === 'number'
+  }
+
+  const areAsyncExchangeFieldsSet =
+    !descriptor?.eservice.asyncExchange ||
+    Boolean(
+      descriptor.asyncExchangeCallbackInterface &&
+      descriptor.asyncExchangeProperties &&
+      hasAsyncExchangeNumber(descriptor.asyncExchangeProperties.responseTime) &&
+      hasAsyncExchangeNumber(descriptor.asyncExchangeProperties.resourceAvailableTime) &&
```
Alepazz: The OpenAPI types mark `responseTime`, `resourceAvailableTime`, `confirmation`, `bulk`, and `maxResultSet` as required on `AsyncExchangeProperties`. The defensive `typeof` checks here only matter if the BE can return a partial `asyncExchangeProperties` object for a draft in progress. Could you confirm whether that's the case? If yes, the schema should probably mark these fields optional so the front end and the type contract are aligned; if no, the checks can be dropped.

src/pages/ProviderEServiceSummaryPage/ProviderEServiceSummary.page.tsx
```diff
@@ -204,6 +214,17 @@ const ProviderEServiceSummaryPage: React.FC = () => {
...
 
+  const isAsyncExchangeValidationLoading =
+    Boolean(descriptor?.eservice.asyncExchange) && isAssociatedKeychainsLoading
+
+  const areAsyncExchangeFieldsSet =
+    !descriptor?.eservice.asyncExchange ||
+    Boolean(
+      descriptor.asyncExchangeCallbackInterface &&
+      descriptor.asyncExchangeProperties &&
+      (associatedKeychains?.pagination?.totalCount ?? associatedKeychains?.results?.length ?? 0) > 0
```
Alepazz: Per `CompactProducerKeychains`, `pagination` is required, so when `associatedKeychains` is defined `pagination.totalCount` is always present. The fallback to `results.length` looks unreachable.

```suggestion
      (associatedKeychains?.pagination.totalCount ?? 0) > 0
```

src/pages/ProviderEServiceSummaryPage/ProviderEServiceSummary.page.tsx
```diff
@@ -462,13 +487,24 @@ const ProviderEServiceSummaryPage: React.FC = () => {
...
+            <Tooltip
+              title={
+                canBePublished()
+                  ? ''
+                  : isAsyncExchangeValidationLoading
+                    ? ''
+                    : arePersonalDataSet
+                      ? t('summary.notPublishableTooltip.label')
+                      : t('summary.missingPersonalDataField')
+              }
```
Alepazz: Three nested ternaries with two branches collapsing to `''`, and the logic partially duplicates the tooltip rules already encoded in `PublishButton`. Could we extract a small helper (or a `ts-pattern` match) shared between the two call sites?

Alepazz (approved):
LGTM. Left a few non-blocking observations inline.

PR #1909, opened 2026-05-14
Feature/PIN-10029 create keychain card in tech spec step

LucaPierfederici:
## 🔗 Issue
[PIN-10029](https://pagopa.atlassian.net/browse/PIN-10029)

## 📝 Description / Context
Add a new "Portachiavi erogatore" card in the "Specifiche tecniche" step of the provider e-service creation flow. The card lets the provider associate one or more existing producer keychains to the e-service being created. It is only relevant for async e-services and is not shown in the template flow (neither when creating a template, nor when instantiating from one).

## 🛠 List of changes
- Add `EServiceProducerKeychainSection` component with a `useFieldArray`-based list of autocomplete rows and an "Aggiungi portachiavi" button to append rows
- Exclude keychains already selected in other rows from each autocomplete's options (via `useWatch` on `keychains`); disable the add button when every available keychain is already selected
- Show a warning alert (and hide the form) when the user is `isOperatorAPI`
- Show a warning alert (and hide the add button) when the organization has no keychains
- Render a read-only `InformationContainer` list of associated keychains (or `-`) when `areEServiceGeneralInfoEditable` is false (i.e. version > 1)
- Integrate the section in `EServiceCreateStepTechSpec`, visible only when the e-service is async, not created from a template and not in template-creation mode
- Split `EServiceCreateStepTechSpec` into an outer container that performs the preload and an inner `EServiceCreateStepTechSpecForm` that owns the form, so the form is mounted only after the initial associated keychains are available
- Preload the keychains already associated to the e-service via `KeychainQueries.getKeychainsList({ eserviceId, limit: 100, offset: 0 })` and show a skeleton while the query is pending
- On step submit, diff initial vs. final selection (reading the latest list from the React Query cache, with a fallback to the preloaded value) and call `addKeychainToEService` / `removeKeychainFromEService` in parallel via `Promise.allSettled`
- On any mutation failure, refetch the associated keychains, reset the `keychains` field to the refreshed server state (preserving the other form values) and abort forward navigation so the user can retry
- Strip `keychains` from the payload used by `compareObjects` so the existing "skip update when nothing changed" short-circuit keeps working
- Add a `hasConfirmationDialog` parameter (default `true`) to `useRemoveKeychainFromEService` so the dialog can be suppressed when removals are triggered as part of the step submit (the explicit removal flow from the keychain detail page keeps the existing confirmation)
- Add i18n keys under `create.step4.producerKeychainSection` (it/en), including `title`, `subtitle`, `keychainField.label`, `addKeychainBtn`, `removeRowTooltip`, `readOnlyLabel`, `noKeychainAlert`, `apiRoleAlert`
- Add unit tests for the new section (visibility, alerts, add/remove rows, exclusion logic, read-only mode) and extend the step tests to cover async / non-async / template visibility, the skeleton during preload, the add/remove diff on submit and the abort-on-failure behavior
- Update e-service mocks with the new `hasProducerKeychain` / `hasProducerKeychainKeys` fields

## 🧪 How to test
- Open the provider e-service creation flow with an async e-service and reach the "Specifiche tecniche" step → the "Portachiavi erogatore" card is visible
- Repeat with a non-async e-service, a template-instance e-service or while creating a template → the card is not shown
- As `isOperatorAPI`, verify only the warning alert is shown and the form is hidden
- With an organization that has no keychains, verify the empty-list alert is shown and the add button is hidden
- Add multiple rows and verify that a keychain selected in one row disappears from the other rows' options, and reappears when deselected/removed
- Select every available keychain and verify the "Aggiungi portachiavi" button is disabled
- Add/remove rows, change selections, then submit the step → verify the associations are persisted (added/removed) before moving forward, and that no mutation is sent for unchanged rows
- Trigger a mutation error (e.g. force a 5xx on add/remove) and verify the step does not advance, the form is repopulated from the server, and the other fields keep their dirty values
- Reopen the step on an e-service version > 1 (non-editable) and verify the read-only list of associated keychains is rendered (or `-` when none)
- Remove a keychain from the e-service detail page (outside this flow) and verify the existing confirmation dialog still appears

Screenshot 2026-05-14 alle 15 20 20
Screenshot 2026-05-14 alle 16 10 58
Screenshot 2026-05-14 alle 16 10 21

[PIN-10029]: https://pagopa.atlassian.net/browse/PIN-10029


src/pages/ProviderEServiceCreatePage/components/EServiceCreateStepTechSpec/EServiceCreateStepTechSpec.tsx
```diff
@@ -37,24 +88,95 @@ export const EServiceCreateStepTechSpec: React.FC<ActiveStepProps> = () => {
...
+      const finalIds = values.keychains
+        .map((row) => row.value?.id)
+        .filter((id): id is string => Boolean(id))
+
+      const addedIds = finalIds.filter((id) => !initialIds.includes(id))
+      const removedIds = initialIds.filter((id) => !finalIds.includes(id))
+
+      if (removedIds.length > 0) {
+        const hasConfirmed = await new Promise((resolve) => {
+          openDialog({
```
borgesis95: `openDialog` with new flow (https://www.figma.com/design/CpRV3kPvFEWLXGtJUgWeZW/Interop-%E2%80%94-Delivery-FE---QA?node-id=7131-50800&t=Zu6qwsffNsDA6RaU-4) is still necessary?

src/pages/ProviderEServiceCreatePage/components/sections/__tests__/EServiceProducerKeychainSection.test.tsx
```diff
@@ -0,0 +1,175 @@
...
+  return { ...actual, useQuery: vi.fn() }
+})
+
+vi.mock('@/api/keychain', () => ({
+  KeychainQueries: {
+    getKeychainsList: vi.fn(() => ({ queryKey: ['KeychainGetList'], queryFn: vi.fn() })),
+  },
+}))
+
+const useJwtSpy = vi.spyOn(AuthHooks, 'useJwt')
```
borgesis95: can we use `mockUseJwt` that is already available?

borgesis95 (commented):
also: remove .`worktrees` folder

src/pages/ProviderEServiceCreatePage/components/EServiceCreateStepTechSpec/EServiceCreateStepTechSpec.tsx
LucaPierfederici: solved in this [commit](https://github.com/pagopa/pdnd-interop-frontend/pull/1909/commits/da2938b9c3a30f0fb1ceeb8a4ed3310a5a22e217)

src/pages/ProviderEServiceCreatePage/components/sections/__tests__/EServiceProducerKeychainSection.test.tsx
LucaPierfederici: solved in this [commit](https://github.com/pagopa/pdnd-interop-frontend/pull/1909/commits/bcf80b375190b50ee3bbeb93446ab9cbb339791e)

borgesis95 (commented):
Take a look on what remains about copilot review

LucaPierfederici:
> Take a look on what remains about copilot review

solved in last commit @borgesis95

src/pages/ProviderEServiceCreatePage/components/sections/EServiceProducerKeychainSection.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+  const { control } = useFormContext<FormValues>()
+  const { fields, append, remove } = useFieldArray<FormValues>({
+    control,
+    name: 'keychains',
+  })
+  const watchedKeychains = useWatch({ control, name: 'keychains' })
+
+  const { data: allKeychains = [], isPending } = useQuery({
+    ...KeychainQueries.getKeychainsList({ limit: 100, offset: 0 }),
```
Alepazz: Why bump this from 50 to 100? Without server-side search/pagination the silent truncation past the limit is still there, just at a higher threshold. Same hardcoded `limit: 100` at `EServiceCreateStepTechSpec.tsx` L53 and L140.

src/api/keychain/keychain.mutations.ts
```diff
@@ -30,7 +30,7 @@ function useCreateKeychain() {
   })
 }
 
-function useRemoveKeychainFromEService() {
+function useRemoveKeychainFromEService(hasConfirmationDialog = true) {
```
Alepazz: Parametrising the hook with a boolean to toggle the confirmation dialog hides two distinct behaviours behind one hook. An alternative that keeps the call site explicit: drop the `confirmationDialog` from this hook's `meta` and let the caller pass it via `mutate(..., { meta: { confirmationDialog: ... } })` when it needs the dialog (or omit it entirely when batching, like the new step submit does). The single-removal flow on the keychain detail page would supply the dialog at the call site, the batch flow wouldn't, and there's no hook flag to inspect.

src/pages/ProviderEServiceCreatePage/components/EServiceCreateStepTechSpec/EServiceCreateStepTechSpec.tsx
```diff
@@ -79,6 +134,57 @@ export const EServiceCreateStepTechSpec: React.FC<ActiveStepProps> = () => {
...
+          const refreshedKeychains = refetched.results
+          formMethods.reset(
+            {
+              ...formMethods.getValues(),
+              keychains:
+                refreshedKeychains.length > 0
+                  ? refreshedKeychains.map((k) => ({ value: k }))
+                  : [{ value: null }],
+            },
+            { keepDirtyValues: false }
```
Alepazz: `keepDirtyValues: false` is the default for `formMethods.reset`, so the entire options argument can be dropped here.

src/pages/ProviderEServiceCreatePage/components/EServiceCreateStepTechSpec/EServiceCreateStepTechSpec.tsx
```diff
@@ -19,11 +21,15 @@ import { trackEvent } from '@/config/tracking'
 import { match } from 'ts-pattern'
 import { EServiceInterfaceSection } from '../sections/EServiceInterfaceSection'
 import { EServiceVoucherSection } from '../sections/EServiceVoucherSection'
+import { EServiceProducerKeychainSection, ProducerKeychainFieldArrayItem } from '../sections/EServiceProducerKeychainSection'
```
Alepazz:
```suggestion
import { EServiceProducerKeychainSection, type ProducerKeychainFieldArrayItem } from '../sections/EServiceProducerKeychainSection'
```
`ProducerKeychainFieldArrayItem` is used only as a type. Inlining the `type` marker keeps the import compliant with TS `verbatimModuleSyntax` and signals intent at a glance.

src/pages/ProviderEServiceCreatePage/components/sections/EServiceProducerKeychainSection.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+export const EServiceProducerKeychainSection: React.FC = () => {
+  const { t } = useTranslation('eservice', {
+    keyPrefix: 'create.step4.producerKeychainSection',
+  })
+  const { isOperatorAPI } = AuthHooks.useJwt()
+  const { areEServiceGeneralInfoEditable } = useEServiceCreateContext()
+
+  const { control } = useFormContext<FormValues>()
+  const { fields, append, remove } = useFieldArray<FormValues>({
```
Alepazz, line 74: In read-only mode (`!areEServiceGeneralInfoEditable`), only `watchedKeychains` is actually consumed: `useFieldArray`, `append`/`remove`, and the `useQuery` for `allKeychains` are all set up but never used. The Rules of Hooks block hoisting the early return, but you could split the section into two components, e.g. a read-only variant and an editable variant, and pick which to render here (or at the parent). It avoids the unused setup and keeps each path tighter.

Alepazz (commented):
A few points before merging, see inline comments.

src/pages/ProviderEServiceCreatePage/components/sections/EServiceProducerKeychainSection.tsx
LucaPierfederici: good point, anyway we don't have unpaginated api call.
also in https://github.com/pagopa/pdnd-interop-frontend/blob/3964f88b0cef402b4fb0ad0e0f9359fa7dcde99b/src/pages/ProviderEServiceDetailsPage/components/ProviderEServiceKeychainsTab/AddKeychainToEServiceDrawer.tsx#L64-L71
This api is used with hardcoded limit.
what do you suggest?

src/api/keychain/keychain.mutations.ts
LucaPierfederici: the `mutate` method doesn't accept `meta` in options parameter, only `onSuccess, onError, onSettled`.
I can edit the function parameter passing an object so it is explicit:
```
function useRemoveKeychainFromEService(options: { withConfirmationDialog?: boolean } = {}) {
  const { withConfirmationDialog = true } = options
  // ...rest
}
```
WDYT?

src/pages/ProviderEServiceCreatePage/components/sections/EServiceProducerKeychainSection.tsx
LucaPierfederici: solved [here](https://github.com/pagopa/pdnd-interop-frontend/pull/1909/commits/ace88e61947108215431bc34227cfc1d0174d17d)

src/pages/ProviderEServiceCreatePage/components/EServiceCreateStepTechSpec/EServiceCreateStepTechSpec.tsx
LucaPierfederici: solved [here](https://github.com/pagopa/pdnd-interop-frontend/pull/1909/commits/6d46143fedbe590a4a04ebf20d15ee4b3d895616)

src/pages/ProviderEServiceCreatePage/components/EServiceCreateStepTechSpec/EServiceCreateStepTechSpec.tsx
LucaPierfederici: solved [here](https://github.com/pagopa/pdnd-interop-frontend/pull/1909/commits/6d46143fedbe590a4a04ebf20d15ee4b3d895616)

src/pages/ProviderEServiceCreatePage/components/EServiceCreateStepTechSpec/EServiceCreateStepTechSpec.tsx
```diff
@@ -34,8 +43,50 @@ export type EServiceCreateStepTechSpecFormValues = {
...
+
+  const { data: initialAssociatedKeychains, isPending } = useQuery({
+    ...KeychainQueries.getKeychainsList({
+      eserviceId: descriptor?.eservice.id ?? '',
+      limit: 100,
+      offset: 0,
+    }),
+    select: (d) => d.results,
+    enabled: isProducerKeychainSectionVisible && Boolean(descriptor?.eservice.id),
+  })
```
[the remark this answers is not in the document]
DenisLantero: @LucaPierfederici Please check this comment, this issue causes the FE to crash when loading the 3rd step in the eservice creation

src/pages/ProviderEServiceCreatePage/components/sections/EServiceProducerKeychainSection.tsx
DenisLantero: Fixed [here](https://github.com/pagopa/pdnd-interop-frontend/pull/1909/commits/44360a50)

src/api/keychain/keychain.mutations.ts
DenisLantero: Fixed [here](https://github.com/pagopa/pdnd-interop-frontend/pull/1909/commits/c738b2dc)

borgesis95 (approved):

src/pages/ProviderEServiceCreatePage/components/EServiceCreateStepTechSpec/__tests__/EServiceCreateStepTechSpec.test.tsx
```diff
@@ -52,8 +89,22 @@ vi.mock('@/api/eservice', () => ({
...
+vi.mock('@/api/keychain', () => ({
+  KeychainQueries: {
+    getAllKeychainsList: vi.fn((params: { eserviceId?: string }) => ({
+      queryKey: ['KeychainGetAllList', params],
+      queryFn: () => Promise.resolve([]),
+    })),
+  },
+  KeychainMutations: {
+    useAddKeychainToEService: () => ({ mutateAsync: addKeychainToEService }),
+    useRemoveKeychainFromEService: () => ({ mutateAsync: removeKeychainFromEService }),
```
Alepazz: The real code now calls `useRemoveKeychainFromEService({ withConfirmationDialog: false })`, but the mock ignores the argument. Tests still pass, but the new options contract is not exercised. Could the mock accept and forward the options (e.g. spy on it) so a future signature change is caught here?

Alepazz (approved):
LGTM. Left a non-blocking suggestion inline.

PR #1921, opened 2026-05-20
feature/PIN-9939: e-service detail page header actions mapping

Alepazz:
## 🔗 Issue
[PIN-9939]

## 📝 Description / Context
This PR maps the new manual-archiving operations onto the provider e-service detail
page header introduced by [PIN-9937] (NewPageContainer redesign) and [PIN-9942]
(versions list modal). It implements the action layout per the Figma spec for the
admin/operatorAPI happy path, while preserving the current legacy behavior for the
delegation, draft-sibling, and template-instance cases. Backend wiring of the
archiving mutations is out of scope and will land with [PIN-9943] / [PIN-9945].

The PR also extends the same slot-split header pattern to the **consumer**
e-service catalog detail page: the existing `Richiedi fruizione` /
`Visualizza richiesta` / `Completa richiesta` actions are promoted from the
3-dots menu (where [PIN-9937] had placed them as a no-thought adapter) into
the proper `primaryAction` / `headerInfoActions` slots, the chip is switched
to the new `for: 'descriptor'` variant, and a new `Vedi ultima versione`
action is exposed in the version sub-header when an obsolete descriptor has
a newer non-draft version available.

## 🛠 List of changes
- **New actions** in `useGetProviderEServiceActions`: `archiveDescriptor`,
  `cancelArchivingDescriptor`, `cancelArchivingEservice`, `archiveEservice`,
  `viewLatestVersion`, `viewAllVersions`. The archiving actions are wired to a
  shared `noop` placeholder (TODO [PIN-9943] / [PIN-9945]); `viewLatestVersion`
  navigates to the latest non-draft descriptor; `viewAllVersions` invokes a
  callback supplied by the caller (the page opens the `EServiceVersionSelectorDrawer`).
- **Slot split** for the details-page happy path (admin/operatorAPI, no delegation,
  no draft sibling, no template instance): exhaustive `ts-pattern` match on
  `(state, archivingScope)` returning `{ primaryAction, headerInfoActions, menuActions }`.
  Other cases fall through to the legacy behavior with `viewAllVersionsAction`
  appended to the menu so version navigation stays reachable.
- **`StatusChip`**: new `for: 'descriptor'` discriminant with feminine Italian labels
  (`attiva`, `sospesa`, `archiviata`, …); `ARCHIVED` uses the MUI default (neutral grey).
- **`NewPageContainer.HeaderInfoSection`**: minor design tweaks
  (`pl={3} pr={0}`, `spacing={1}` between inline actions).
- **`ProviderEServiceGeneralInfoSection`**: removed the inline
  "Naviga a un'altra versione dell'e-service" bottom action and the local drawer
  state; the drawer is now lifted to the page level and opened via the new
  "Vedi tutte le versioni" menu item.
- **`ProviderEServiceDetails.page.tsx`**: wires the new params (`where`,
  `archivingSchedule` from `descriptor?.archivingSchedule`, `latestDescriptorId`,
  `onViewAllVersions`, `hasPersonalData` from `descriptor?.eservice.personalData`),
  renders the `EServiceVersionSelectorDrawer`, and uses the shared
  `getLastDescriptor` utility for the latest-version computation.
- **i18n**: new keys under `eservice.read.actions.*`
  (`suspendVersion`, `reactivateVersion`, `createNewVersion`, `archiveVersion`,
  `cloneEservice`, `archiveEservice`, `cancelArchivingVersion`,
  `cancelArchivingEservice`, `viewLatestVersion`, `viewAllVersions`) and the
  missing `common.status.eservice.{ARCHIVING, ARCHIVING_SUSPENDED}`
  entries (cleans up a typecheck error left as a placeholder by an earlier
  [commit][placeholder-commit] on the [PIN-9936] feature branch).
- **Consumer-side changes** (`useGetEServiceConsumerActions`,
  `ConsumerEServiceDetails.page.tsx`, `ConsumerEServiceGeneralInfoSection.tsx`):
  hook refactored to return
  `{ primaryAction, secondaryAction, menuActions, headerInfoActions }` instead
  of bundling everything in `menuActions`. Primary slot picks one of
  `inspect` / `editDraft` / `subscribe` / `subscribeDisabledMissingAttrs` with
  priority order `inspect > editDraft > subscribe`. All primary candidates
  use `variant: 'contained'` to match the Figma. The new `viewLatestVersion`
  action lives in `headerInfoActions` (sub-header right side) when the user
  is on an obsolete descriptor and a newer non-draft version exists. The
  chip is switched to `for: 'descriptor'` (feminine Italian labels). The
  legacy inline bottom-action `Naviga a un'altra versione dell'e-service`
  is removed from the `Informazioni generali` section; the existing
  version-list modal (chip number clickable) remains as the entry point for
  version navigation.
- **Shared util**: `getViewLatestVersionTargetId` extracted to
  `src/utils/eservice.utils.ts` and reused by both provider and consumer
  pages.
- **Fix** in `NewPageContainer.ActionsSection`: the destructured `disabled`
  prop of `ActionItemButton` was never forwarded to `<Button>`. Without
  this fix a disabled `primaryAction` (e.g. `Richiedi fruizione` disabled
  with the missing-certified-attributes tooltip) rendered as a fully
  enabled clickable button. Pre-existing bug from [PIN-9937].
- **i18n cleanup**: removed orphan key
  `eservice.read.sections.generalInformations.bottomActions.navigateVersions`
  (it/en) after dropping the inline navigation action.
- **Tests**: 75 tests for `useGetProviderEServiceActions`, including new coverage
  for the slot split (happy path) and the legacy bypass. 16 tests for
  `useGetEServiceConsumerActions` adapted to the new return shape, plus 2
  new tests covering the `viewLatestVersion` headerInfoAction (presence
  and navigation).

## ⚠️ Drive-by fixes (pre-existing bugs surfaced during review)

Two pre-existing bugs in `src/utils/eservice.utils.ts` are fixed here, because this
PR adds two new consumers for both utilities (`viewLatestVersion` on provider and
consumer) and would have amplified their reachability.

> 🐛 **`getLastDescriptor`: lexicographic vs numeric version comparison**
> The utility compared `version: string` with `>=` directly, which is a
> **lexicographic** comparison: `'10' < '2'` because `'1'` sorts before `'2'`. Any
> e-service with 10+ versions would have returned the wrong "latest descriptor".
> Fixed by switching to `Number(a.version) >= Number(b.version)` numeric
> comparison. A dedicated 9-vs-10 test is added to lock the behavior.

> 🐛 **`getViewLatestVersionTargetId`: did not exclude `ARCHIVED`**
> The utility filtered out `DRAFT` and `WAITING_FOR_APPROVAL` but **not**
> `ARCHIVED`. So on an obsolete descriptor whose only newer sibling was itself
> archived, "Vedi ultima versione" would have navigated to an archived
> descriptor. The sibling utility `getActiveDescriptor` already excluded
> `ARCHIVED`; the inconsistency between the two was the original cause. Fixed by
> aligning the filter. Test coverage added.

## 📐 Design decisions
- **Menu actions on non-latest versions** (Figma confirmation in [this thread](https://www.figma.com/design/CpRV3kPvFEWLXGtJUgWeZW?node-id=5281-10822#1768772517)):
  `Crea nuova versione` and `Archivia e-service` are intentionally kept in the
  3-dots menu of every version, not only the latest one. The product expectation
  is that the operator can still open a new draft or archive the e-service while
  inspecting a previous version.
- **Two entry points to the versions list coexist by design**: clicking the
  version number in the header info row opens the modal `showEserviceVersionsList`
  (added by [PIN-9942]); clicking `Vedi tutte le versioni` from the 3-dots menu
  opens the side drawer (`EServiceVersionSelectorDrawer`). Both paradigms are
  present in the Figma — see [this frame](https://www.figma.com/design/CpRV3kPvFEWLXGtJUgWeZW/Interop-%E2%80%94-Delivery-FE---QA?node-id=5281-10968&t=ftYMNqAwKEbAhgYB-4)
  showing the two surfaces side by side. They can be unified in a separate PR
  if product decides only one should remain.
- **No `Aggiorna a nuova versione` shortcut on the consumer catalog detail**:
  the Figma initially included this action on obsolete descriptors with an
  upgradable agreement. Confirmed with design ([Figma conversation](https://www.figma.com/design/CpRV3kPvFEWLXGtJUgWeZW?node-id=5430-18592#1765949282))
  that the existing flow is already optimal: from the obsolete catalog page,
  `Visualizza richiesta` lands the user directly on the agreement detail
  (or opens a tenant selector when the operator represents multiple consumer
  organizations on the same e-service), where `Aggiorna a nuova versione`
  already lives. Two clicks total. No new wiring needed in this PR; the
  multi-tenant delegation UX may be revisited in a separate ticket if/when
  design runs the deepdive Fabio mentioned.
- **`Duplica e-service`** label: per Figma the provider e-service menu must show
  `Duplica e-service`, not the generic `Clona`/`Duplica` used elsewhere. A new
  dedicated i18n key `eservice.read.actions.cloneEservice` was introduced to
  avoid hijacking the shared `common.actions.clone` (used by agreement and
  purpose menus). Note: the screenshots above still display the previous
  `Clona` label — see the warning callout next to the screenshots.

## 🚩 Open points to keep in mind
- **Archiving mutations**: `archiveDescriptor`, `archiveEservice`, and the two
  `cancelArchiving*` actions are wired to a `noop`; the actual dialogs and BFF
  calls will land in [PIN-9943] (descriptor-level archiving dialog) and
  [PIN-9945] (e-service-level archiving dialog).
## 🧪 How to test
- On the provider e-service detail page (`/erogazione/e-service/:id/:descriptorId`),
  open an e-service whose current descriptor is **active** with no draft sibling,
  not a template instance, not in delegation:
  - the version row shows `Sospendi versione` and `Crea nuova versione` inline;
  - the 3-dots menu contains `Duplica e-service`, `Archivia e-service`, `Vedi tutte le versioni`.
- Repeat on a **deprecated** version: header shows `Sospendi versione` and
  `Archivia versione`; menu has `Crea nuova versione`, `Duplica e-service`,
  `Archivia e-service`, `Vedi tutte le versioni`.
- On a **suspended** version: header shows `Riattiva versione` and
  `Archivia versione`; menu has `Duplica e-service`, `Crea nuova versione`,
  `Vedi tutte le versioni`.
- On an **archived** version where a newer version exists: header shows
  `Vedi ultima versione`; menu has `Duplica e-service`, `Archivia e-service`,
  `Vedi tutte le versioni`.
- For the archiving-overlay states (state `ARCHIVING` / `ARCHIVING_SUSPENDED`
  with scope `DESCRIPTOR` or `ESERVICE`): not reachable from the dev backend
  yet (the BFF deploy of the archivingSchedule data is not on dev/uat yet); see
  the screenshots attached by the author for the expected layouts.
- For legacy bypass cases (template instance, draft sibling, delegations): the
  header behaviour matches what is on `dev` today, with `Vedi tutte le versioni`
  appended to the 3-dots menu.
- Click `Vedi tutte le versioni` and confirm the side drawer
  `Vedi un'altra versione dell'e-service` opens with the version selector and
  the `Vai alla versione selezionata` CTA.
- Verify the inline link `Naviga a un'altra versione dell'e-service` is no
  longer present in the `Informazioni generali` section.

### Consumer (`/fruizione/catalogo-e-service/:eserviceId/:descriptorId`)

- **Active descriptor, no agreement, certified attributes OK**:
  `Richiedi fruizione` contained blue primary (enabled).
- **Active descriptor, no agreement, missing certified attributes**:
  `Richiedi fruizione` contained but disabled + tooltip
  "Il tuo ente non ha gli attributi certificati necessari per iscriversi".
- **Suspended descriptor, no agreement**: `Richiedi fruizione` disabled +
  tooltip "Non è attualmente possibile presentare nuove richieste,
  l'e-service è sospeso".
- **Active descriptor, active / suspended / pending agreement**:
  `Visualizza richiesta` contained primary.
- **DRAFT agreement (form started, not submitted)**: `Completa richiesta`
  contained primary (replaces `Richiedi fruizione`). A real example on dev
  is the e-service `pozzallo eservice delega in erogazione` of
  `Comune di Pozzallo` (descriptor v.1).
- **Obsolete descriptor with active agreement and a newer non-draft
  version**: `Visualizza richiesta` contained primary on the header row,
  `Vedi ultima versione` (text + `ReplayCircleFilled` icon) in the
  sub-header right side, chip shows "obsoleta".
- **Active descriptor, the user represents multiple organizations (own +
  at least one consumer delegator on the same e-service), one of them
  has an active / suspended / pending agreement and another has no
  agreement**: `Visualizza richiesta` contained primary + `Richiedi fruizione`
  outlined secondary inline on the right. Triggering the secondary opens
  the `DialogSelectAgreementConsumer` so the user can pick which
  represented organization should subscribe.
  ([Figma conversation](https://www.figma.com/design/CpRV3kPvFEWLXGtJUgWeZW?node-id=5446-19417#1766189419))
- **Same multi-tenant delegation scenario but the existing agreement is
  in DRAFT** (one represented organization started but did not submit the
  request): `Completa richiesta` contained primary + `Richiedi fruizione`
  outlined secondary.
- **Same multi-tenant delegation scenario on a SUSPENDED descriptor**:
  `Richiedi fruizione` is still rendered as outlined secondary but
  disabled, with the usual `Non è attualmente possibile presentare nuove
  richieste, l'e-service è sospeso` tooltip.
- The bottom-action `Naviga a un'altra versione dell'e-service` is no
  longer rendered inside the `Informazioni generali` section; navigation
  between versions remains via the clickable version number that opens
  the `Stato versioni` modal.

---

> :warning: **Screenshots below show `Clona` in the 3-dots menu.** The actual
> label rendered by this branch is now `Duplica e-service` (and `Duplicate
> e-service` in English): the clone action uses a new dedicated i18n key
> (`eservice.read.actions.cloneEservice`) instead of the shared
> `common.actions.clone`, so the agreement and purpose menus keep saying
> `Clona` while only the provider e-service menu shows `Duplica e-service` —
> per Figma spec linked [here](https://www.figma.com/design/CpRV3kPvFEWLXGtJUgWeZW?node-id=5281-10822#1767976150).
> Screenshots have not been regenerated to save time.

### Versione attiva (PUBLISHED) — header con bottoni inline "Sospendi versione" e "Crea nuova versione"
01-published

### Versione attiva (PUBLISHED) — menu 3-puntini aperto: Clona, Archivia e-service, Vedi tutte le versioni
01-published

### Versione obsoleta (DEPRECATED) — header con bottoni inline "Sospendi versione" e "Archivia versione"
02-deprecated

### Versione obsoleta (DEPRECATED) — menu 3-puntini aperto: Crea nuova versione, Clona, Archivia e-service, Vedi tutte le versioni
02-deprecated-menu

### Versione sospesa (SUSPENDED) — header con bottoni inline "Riattiva versione" e "Archivia versione"
03-suspended-active-descriptor-no-menu

### Versione sospesa (SUSPENDED) — menu 3-puntini aperto: Clona, Crea nuova versione, Archivia e-service, Vedi tutte le versioni
03-suspended-active-descriptor

### Versione archiviata (ARCHIVED) con versione più recente esistente — header con "Vedi ultima versione" inline, chip grigio neutro "archiviata"
04-archived

### Versione archiviata (ARCHIVED) — menu 3-puntini aperto: Clona
02-archived-menu-only-clone

### Versione in archiviazione (ARCHIVING, scope DESCRIPTOR) — header con "Sospendi versione" e "Annulla archiviazione versione" inline
05-archiving-descriptor

### Versione in archiviazione (ARCHIVING, scope DESCRIPTOR) — menu 3-puntini aperto: Crea nuova versione, Clona, Archivia e-service, Vedi tutte le versioni
05-archiving-descriptor-menu

### Versione sospesa e in archiviazione (ARCHIVING_SUSPENDED, scope DESCRIPTOR) — header con "Riattiva versione" e "Annulla archiviazione versione" inline
06-archiving-suspended-descriptor

### Versione sospesa e in archiviazione (ARCHIVING_SUSPENDED, scope DESCRIPTOR) — menu 3-puntini aperto: Crea nuova versione, Clona, Archivia e-service, Vedi tutte le versioni
06-archiving-suspended-descriptor-menu

### Versione attiva con intero e-service in archiviazione (PUBLISHED, scope ESERVICE) — primary contained "Annulla archiviazione e-service" in alto, "Sospendi versione" inline
07-published-eservice-archiving-overlay

### Versione attiva con intero e-service in archiviazione (PUBLISHED, scope ESERVICE) — menu 3-puntini aperto: Clona, Vedi tutte le versioni
07-published-eservice-archiving-overlay-menu

### Versione sospesa con intero e-service in archiviazione (SUSPENDED, scope ESERVICE) — primary contained "Annulla archiviazione e-service" in alto, "Riattiva versione" inline
08-suspended-eservice-archiving-overlay

### Versione sospesa con intero e-service in archiviazione (SUSPENDED, scope ESERVICE) — menu 3-puntini aperto: Clona, Vedi tutte le versioni
08-suspended-eservice-archiving-overlay-menu

---
---
### Versione attiva (PUBLISHED), nessuna richiesta di fruizione e attributi certificati posseduti — primary contained "Richiedi fruizione" abilitato, chip "attiva"
01-published-can-subscribe

  ### Versione attiva (PUBLISHED), nessuna richiesta di fruizione, attributi certificati mancanti — "Richiedi fruizione" contained ma disabilitato con tooltip "Il tuo ente non ha gli attributi certificati necessari per iscriversi"
02-published-missing-attrs-tooltip

  ### Versione sospesa (SUSPENDED), nessuna richiesta di fruizione — "Richiedi fruizione" disabilitato con tooltip "Non è attualmente possibile presentare nuove richieste, l'e-service è sospeso", chip "sospesa"
03-suspended-no-agreement

  ### Versione obsoleta (DEPRECATED) con richiesta di fruizione attiva e versione più recente disponibile — primary contained "Visualizza richiesta", "Vedi ultima versione" inline nel sub-header, chip "obsoleta"
04-deprecated-with-active-agreement-and-latest

  ### Versione attiva (PUBLISHED) con richiesta di fruizione attiva — primary contained "Visualizza richiesta", chip "attiva"
05-published-with-active-agreement

  ### Versione attiva (PUBLISHED) con richiesta di fruizione in bozza (DRAFT) non sottomessa — primary contained "Completa richiesta", chip "attiva"
06-published-with-draft-agreement

### Versione attiva (PUBLISHED), operatore che rappresenta più organizzazioni (propria + delegante) con bozza (DRAFT) su una e nessuna richiesta sull'altra — primary contained "Completa richiesta" + outlined secondary "Richiedi fruizione", chip "attiva"
E-editdraft-plus-subscribe

### Versione sospesa (SUSPENDED), operatore che rappresenta più organizzazioni con fruizione attiva su una e nessuna sull'altra — primary contained "Visualizza richiesta" + outlined secondary "Richiedi fruizione" disabilitato con tooltip "Non è attualmente possibile presentare nuove richieste, l'e-service è sospeso", chip "sospesa"
F2-suspended-tooltip-on-hover

[PIN-9936]: https://pagopa.atlassian.net/browse/PIN-9936
[PIN-9937]: https://pagopa.atlassian.net/browse/PIN-9937
[PIN-9939]: https://pagopa.atlassian.net/browse/PIN-9939
[PIN-9942]: https://pagopa.atlassian.net/browse/PIN-9942
[PIN-9943]: https://pagopa.atlassian.net/browse/PIN-9943
[PIN-9945]: https://pagopa.atlassian.net/browse/PIN-9945
[PIN-10054]: https://pagopa.atlassian.net/browse/PIN-10054
[clone-rename-commit]: https://github.com/pagopa/pdnd-interop-frontend/commit/419965f2
[placeholder-commit]: https://github.com/pagopa/pdnd-interop-frontend/commit/3b6dd2e9
[be-pr-3335]: https://github.com/pagopa/interop-be-monorepo/pull/3335


src/components/shared/StatusChip.tsx
```diff
@@ -157,6 +168,16 @@ export const StatusChip: React.FC<StatusChipProps> = (props) => {
       : t(`status.eservice.${props.state}`)
   }
 
+  if (props.for === 'descriptor') {
```
alten-dturus, line 171: Shouldn't this logic also be applied to `ARCHIVING-SUSPENDED`?

src/hooks/useGetProviderEServiceActions.ts
```diff
@@ -151,10 +159,59 @@ export function useGetProviderEServiceActions(
...
+    icon: CancelOutlinedIcon,
+  }
+
+  const cancelArchivingEserviceAction: ActionItemButton = {
+    action: noopArchivingAction,
+    label: tEserviceActions('cancelArchivingEservice'),
+    icon: CancelOutlinedIcon,
+    variant: 'contained',
+  }
+
```
alten-dturus, line 196: Only for an aesthetic question I prefer that all archive actions are in the same order nearly one to another and then the view version actions. Like:
archive descriptor
cancel archiving descriptor
archive eservice
cancel archiving eservice
view latest version
view all versions

src/hooks/useGetProviderEServiceActions.ts
```diff
@@ -1018,10 +1069,94 @@ export function useGetProviderEServiceActions(
...
+      primary: undefined,
+      header: [suspendAction, createNewDraftAction],
+      menu: menuClassic,
+    }))
+    .with({ state: 'DEPRECATED' }, () => ({
+      primary: undefined,
+      header: [suspendAction, archiveDescriptorAction],
+      menu: menuWithNewVersion,
+    }))
+    .with({ state: 'SUSPENDED' }, () => ({
```
alten-dturus, line 1125: I'm not sure this is correct. I think that the `SUSPENDED` state should has the same structure of the `PUBLISHED` with only the difference that the `suspendAction` is substituted by the `reactivateVersion` because the `SUSPENDED` can be applied only to the last descriptor and not to the deprecated ones. Why put the `archiveDescriptorAction` too?

src/hooks/useGetProviderEServiceActions.ts
```diff
@@ -1018,10 +1069,94 @@ export function useGetProviderEServiceActions(
...
+      primary: undefined,
+      header: [suspendAction, cancelArchivingDescriptorAction],
+      menu: menuWithNewVersion,
+    }))
+    .with({ state: 'ARCHIVING_SUSPENDED', archivingScope: 'ESERVICE' }, () => ({
+      primary: cancelArchivingEserviceAction,
+      header: [reactivateAction],
+      menu: menuEserviceArchiving,
+    }))
+    .with({ state: 'ARCHIVING_SUSPENDED' }, () => ({
```
alten-dturus, line 1158: Are we sure this is a real possibility? I knew that `ARCHIVING_SUSPENDED` will be a state that only the last descriptor can reach because i didn't know that `DEPRECATED` descriptor can be `SUSPENDED` because suspend and reactivate were action only for the active descriptor. The reactivate action name is literally publish if I'm not in error

alten-dturus (commented):
I added some comments

src/components/shared/StatusChip.tsx
Alepazz: Good catch, the remap should apply to `ARCHIVING_SUSPENDED` too for symmetry. Visually nothing changes today since the `ARCHIVING_SUSPENDED` label and color already match `SUSPENDED`, but the explicit remap makes the intent robust against future label/color changes. Will fix.

src/hooks/useGetProviderEServiceActions.ts
Alepazz: The `archiveDescriptorAction` here is intentional. In the manual-archiving design, the single-version archive affordance is enabled exactly when the descriptor is `SUSPENDED`, not when it's `PUBLISHED`. So `SUSPENDED` differs from `PUBLISHED` not only by _reactivate_ vs _suspend_, but also by exposing _Archivia versione_ in the header, as you can check [here](https://www.figma.com/design/CpRV3kPvFEWLXGtJUgWeZW/Interop-%E2%80%94-Delivery-FE---QA?node-id=5975-22464&t=khDpqOtKK8gHFUnE-4).

src/hooks/useGetProviderEServiceActions.ts
Alepazz: This case is reachable. A latest descriptor in `PUBLISHED` can be _suspended_ and then put in single-version manual archiving, ending up in `ARCHIVING_SUSPENDED` with `scope = 'DESCRIPTOR'`. The expected layout (chip sospesa, scheduled-archival badge, header Riattiva versione + Annulla archiviazione versione) is the one shown in the screenshot Versione sospesa e in archiviazione (`ARCHIVING_SUSPENDED`, scope `DESCRIPTOR`) in the PR description.

src/hooks/useGetProviderEServiceActions.ts
alten-dturus: But if we are in the last descriptor we cannot archive it without archive the entire eservice so I think the design is wrong exposing a case that is not correct

src/hooks/useGetProviderEServiceActions.ts
alten-dturus: Same response as [here](https://github.com/pagopa/pdnd-interop-frontend/pull/1921#discussion_r3298844583). The last descriptor cannot be archived manually without archiving the entire eservice. The version archiving is possible only for deprecated versions

src/hooks/useGetProviderEServiceActions.ts
Alepazz: Solved.

src/hooks/useGetProviderEServiceActions.ts
Alepazz: You're right! Fixed.

src/components/shared/StatusChip.tsx
```diff
@@ -157,6 +168,21 @@ export const StatusChip: React.FC<StatusChipProps> = (props) => {
...
+    const isActiveDescriptorBeingArchived =
+      props.isActiveDescriptor &&
+      (props.state === 'ARCHIVING' || props.state === 'ARCHIVING_SUSPENDED')
+
+    const remappedState: EServiceDescriptorState = isActiveDescriptorBeingArchived
```
borgesis95, line 176: Do you think is too much? 
```
function getDisplayState(state: EServiceDescriptorState, isBeingArchived: boolean): EServiceDescriptorState {
  if (!isBeingArchived) return state
  return state === 'ARCHIVING' ? 'PUBLISHED' : 'SUSPENDED'
}
```

```suggestion
   const remappedState = getDisplayState(props.state, isActiveDescriptorBeingArchived)
```

src/hooks/useGetEServiceConsumerActions.ts
```diff
@@ -39,15 +42,14 @@ function useGetEServiceConsumerActions(
   const { mutate: createAgreementDraft } = AgreementMutations.useCreateDraft()
   const { mutate: submitToOwnEService } = AgreementMutations.useSubmitToOwnEService()
 
-  const actions: Array<ActionItemButton> = []
+  const emptyResult = {
```
borgesis95, line 45:
```suggestion
  const emptyActionsList = {
```

borgesis95 (approved):
LGTM, just two points: 
- "Vedi tutte le versioni" doesn't work if you have only one version 
- about testing of hooks (maybe we had this condition before): I don't like to test where the button is e.g (`PUBLISHED: suspend and createNewVersion in header, clone+archiveEservice+viewAllVersions in menu, no primary`). In the hook I'd like to test something like this:
If state is "PUBLISHED  then actions returned will be: _[CREATE,CLONE,ARCHIVE., etc..]._ Testing about where the action is should be something strictly releated to page . What do you think about it? (We can open a mixed activities and discussion for this)

src/components/shared/StatusChip.tsx
Alepazz: Good point on the nested ternary. Refactored using match from `ts-pattern` instead of a standalone `getDisplayState` helper, to stay consistent with the project convention of using` ts-pattern` for state-based branching. The remap reads end-to-end as a single expression now:
  
```
const remappedState: EServiceDescriptorState = match({
     state: props.state,
     isActiveDescriptorBeingArchived,
   })
     .with({ isActiveDescriptorBeingArchived: false }, ({ state }) => state)
     .with({ state: 'ARCHIVING' }, () => 'PUBLISHED' as const)
     .otherwise(() => 'SUSPENDED' as const)
```

src/hooks/useGetEServiceConsumerActions.ts
Alepazz: I'd keep `emptyResult` here: the returned object isn't just an actions list, it bundles `primaryAction`, `secondaryAction`, `menuActions` and `headerInfoActions`. `emptyResult` reflects the full hook return shape; `emptyActionsList` would mislead by suggesting only the menu list. Happy to revisit if you prefer a different name (e.g. emptyActions), but I'd avoid the `…List` suffix.

alten-dturus (approved):

Alepazz:
@borgesis95 Thanks for the thought, but I think keeping the slot in the test makes sense here, because in this case the slot is the spec, not a rendering detail.
Take _Sospendi versione_ as an example: in `PUBLISHED` it sits inline in the header next to _Crea nuova versione_; in _PUBLISHED + ESERVICE archiving_ overlay it's still inline but alone (`createNewVersion` disappears); in _ARCHIVING DESCRIPTOR_ it sits inline next to _Annulla archiviazione_; in _ARCHIVING + DESCRIPTOR + isActiveDescriptor_ it disappears entirely. Same action, different slot and different companions depending on the state. That mapping is not "where the page chooses to render it": it's defined per state in the Figma.
Concrete failure mode: if a regression moved _Annulla archiviazione e-service_ from the primary slot (big blue CTA at the top) to the header inline next to _Sospendi_, the visual change is huge but the action list returned by the hook would be the same. A test asserting only the list would pass; the current test would catch it.

So I'd rather keep the slot assertions where they are. If we did move to "hook returns flat actions + page decides slot", the page would have to own the same (state, scope, isActive) → slot state machine, duplicated for provider and consumer — which is what the hook is doing today. Happy to revisit as a separate refactor if we want to challenge that, but I'd keep this PR as is.

PR #1935, opened 2026-05-27
feature: refactor attribute component (PIN-10101)

alten-dturus:
## 🔗 Issue

[PIN-10101](https://pagopa.atlassian.net/browse/PIN-10101)

## 📝 Description / Context

This PR implement the new design for attribute group components (it adds the certified discrete attribute kind) that include the new details drawer and the new configure certified discrete attribute drawer.
The design is reachable at this [link](https://www.figma.com/design/CpRV3kPvFEWLXGtJUgWeZW/Interop-%E2%80%94-Delivery-FE---QA?node-id=6515-115488&t=LeDjm6A4m6gemgpe-0)

## 🛠 List of changes

- added feature flag for certified discrete attribute
- update definition of `AttributeKey`
- update `generatedTypes`
- refactor `AttrbuteContainer `and `AttributeAutocomplete`
- modified `Drawer` shared component
- implement `ConfigureCertifiedDiscreteAttributeDrawer` and added to `AttributeGroup`

## 🧪 How to test

The components are not testable until the BE integration will be completed

[PIN-10101]: https://pagopa.atlassian.net/browse/PIN-10101?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/shared/AttributeAutocomplete.tsx
```diff
@@ -48,9 +54,19 @@ export const AttributeAutocomplete: React.FC<AttributeAutocompleteProps> = ({
     return result
   }
 
+  const kindsFilter = match(attributeKey)
+    .returnType<Array<AttributeKind>>()
+    .with('certified', () => {
+      if (areCertifiedDiscreteOptionsIncluded) return ['CERTIFIED', 'CERTIFIED_DISCRETE']
```
borgesis95, line 62: Do we need to check here also if FEATURE_FLAG is enabled?

src/components/layout/containers/AttributeContainer.tsx
```diff
@@ -48,138 +66,178 @@ export const AttributeContainer = <
...
+  const [isDetailsDrawerOpen, setIsDetailsDrawerOpen] = React.useState<boolean>(false)
+  const { data: completeAttribute, isLoading: isLoadingCompleteAttribute } = useQuery(
+    AttributeQueries.getSingle(attribute.id)
+  )
+
+  const getMenuActions = () => {
+    const actions: Array<ActionItemButton> = []
+
+    if (
+      FEATURE_FLAG_CERTIFIED_ATTRIBUTE_DISCRETE &&
```
borgesis95: put feature flag and KIND on separated var (as this is used several times)

src/components/shared/AddAttributesToForm/AttributeGroup.tsx
```diff
@@ -57,80 +72,133 @@ export const AttributeGroup: React.FC<AttributeGroupProps> = ({
...
+    const discreteConfig: EServiceAttributeCertifiedDiscreteConfig = {
+      comparator: comparator,
+      threshold: threshold,
+    }
+
+    if (groups[groupIndex].some((att) => att.id === attribute.id)) {
+      groups[groupIndex] = group.map((att) =>
+        att.id === attribute.id ? { ...att, discreteConfig: discreteConfig } : att
+      )
+    } else {
```
borgesis95, line 95: Maybe here is better to create a copy of array ? That's avoid direct mutation of groups var

src/components/shared/AddAttributesToForm/ConfigureCertifiedDiscreteAttributeDrawer.tsx
```diff
@@ -0,0 +1,144 @@
...
+  React.useEffect(() => {
+    if (isOpen) {
+      formMethods.reset({
+        comparator: 'GT',
+        threshold: undefined,
+      })
+    }
+  }, [isOpen, formMethods, attribute])
+
+  const ALL_COMPARATORS: AttributeCertifiedDiscreteComparator[] = [
```
borgesis95: can we move this into `constants` file? Or outside components

src/components/shared/AddAttributesToForm/ConfigureCertifiedDiscreteAttributeDrawer.tsx
```diff
@@ -0,0 +1,144 @@
...
+  })
+  const { isOpen, close, attribute } = useConfigureCertifiedDiscreteAttributeDrawer()
+  const formMethods = useForm<ConfigureCertifiedDiscreteAttributeFormValues>({
+    defaultValues: {
+      comparator: attribute?.discreteConfig?.comparator ?? 'GT',
+      threshold: attribute?.discreteConfig?.threshold ?? undefined,
+    },
+  })
+
+  const handleFormSubmit: SubmitHandler<ConfigureCertifiedDiscreteAttributeFormValues> = ({
```
borgesis95, line 59:
```suggestion
  const handleSubmit: SubmitHandler<ConfigureCertifiedDiscreteAttributeFormValues> = ({
```

borgesis95 (commented):
just several comment (I know that is in DRAFT yet)

borgesis95:
Name of feature flag should be equal to what is available on BE: should be FEATURE_FLAG_ATTRIBUTE_CERTIFIED_DISCRETE

src/components/shared/AddAttributesToForm/ConfigureCertifiedDiscreteAttributeDrawer.tsx
alten-dturus: When it will be used we will have formMethods.handleSubmit(handleSubmit). In this way we have different names

src/components/shared/AddAttributesToForm/ConfigureCertifiedDiscreteAttributeDrawer.tsx
alten-dturus: Yes, we can but they are not used in any other place. It is a const used only to map in this function. Maybe i can do it manually without the map function and the comparators const

src/components/shared/AttributeAutocomplete.tsx
alten-dturus: I don't think because it's optional and default is false so to use it we have to specifically want to do this. Maybe outside I can put the presence of the FLAG as a param that can make this prop passed true

borgesis95 (approved):

src/components/shared/AddAttributesToForm/AttributeGroup.tsx
```diff
@@ -57,80 +72,133 @@ export const AttributeGroup: React.FC<AttributeGroupProps> = ({
...
+    const group = groups[groupIndex]
+
+    const discreteConfig: EServiceAttributeCertifiedDiscreteConfig = {
+      comparator: comparator,
+      threshold: threshold,
+    }
+
+    if (groups[groupIndex].some((att) => att.id === attribute.id)) {
+      groups[groupIndex] = group.map((att) =>
+        att.id === attribute.id ? { ...att, discreteConfig: discreteConfig } : att
```
DenisLantero, line 93: I think this config won’t survive the save. Here we add `discreteConfig` to the descriptor attribute in form state, but the payload mapper still only forwards `id`, `explicitAttributeVerification`, and `dailyCallsPerConsumer`: https://github.com/pagopa/pdnd-interop-frontend/blob/4044d0e3032c3d5fd0cb6228d36b6bf653d50525/src/utils/attribute.utils.ts#L255-L259

Could we also include `discreteConfig` in that mapper and maybe add a focused util test for it?

src/components/shared/AddAttributesToForm/AttributeGroup.tsx
```diff
@@ -57,80 +72,133 @@ export const AttributeGroup: React.FC<AttributeGroupProps> = ({
...
+            )}
+          </>
+        )}
+      </AttributeGroupContainer>
+      {FEATURE_FLAG_ATTRIBUTE_CERTIFIED_DISCRETE && attributeKey === 'certified' && (
+        <ConfigureCertifiedDiscreteAttributeDrawer
+          onSubmit={handleSubmitConfigureDiscreteAttributeDrawer}
+          submitButtonLabel={t('add')}
+        />
       )}
```
DenisLantero, line 203: This looks risky with multiple certified groups. Each `AttributeGroup` renders its own configure drawer, but all drawers share the same zustand state. Opening one discrete attribute opens every drawer instance, and each submit handler closes over a different `groupIndex`, so the config can be written to the wrong group at https://github.com/pagopa/pdnd-interop-frontend/blob/4044d0e3032c3d5fd0cb6228d36b6bf653d50525/src/components/shared/AddAttributesToForm/AttributeGroup.tsx#L90-L96

Could we either render a single drawer higher up, or store both `attribute` and `groupIndex` in the drawer state?

src/components/layout/containers/AttributeContainer.tsx
```diff
@@ -48,138 +66,175 @@ export const AttributeContainer = <
...
+  onOpenConfigDrawer,
 }: AttributeContainerProps<TAttribute>) => {
   const { t } = useTranslation('shared-components', { keyPrefix: 'attributeContainer' })
-  const panelContentId = React.useId()
-  const headerId = React.useId()
-  const alreadyPrefetched = React.useRef(false)
-  const [hasExpandedOnce, setHasExpandedOnce] = React.useState(false)
+  const { t: tCommon } = useTranslation('common', { keyPrefix: 'comparators' })
+  const [isDetailsDrawerOpen, setIsDetailsDrawerOpen] = React.useState<boolean>(false)
+  const { data: completeAttribute, isLoading: isLoadingCompleteAttribute } = useQuery(
```
DenisLantero, line 74: Non-blocking: I noticed this now fetches the full attribute details as soon as each card renders, even though those details are only used when opening the “inspect details” drawer.

Probably fine if these lists stay small, but since `AttributeContainer` is also used in mapped attribute lists, we could consider enabling this query only when `isDetailsDrawerOpen` is true.

alten-dturus: This is like this because before it was done with a prefetch and then a fetch that takes data from cache when the accordion was opened. Now I choose to use that approach so we don't have any loading in the drawer

DenisLantero: Ok perfect, we should keep it like this then, thanks!

src/components/shared/AddAttributesToForm/AttributeGroup.tsx
alten-dturus: The remap function does not have any test before because it is only an assignment, do you think I should add any?

src/components/shared/AddAttributesToForm/AttributeGroup.tsx
```diff
@@ -57,80 +73,134 @@ export const AttributeGroup: React.FC<AttributeGroupProps> = ({
     setIsAttributeAutocompleteVisible(false)
   }
 
+  const handleSubmitConfigureDiscreteAttributeDrawer = (
+    comparator: AttributeCertifiedDiscreteComparator,
+    threshold: number
+  ) => {
+    if (!attribute || !attributeGroupIndex) return
```
Alepazz: This guard returns early when `attributeGroupIndex` is `0`, but `0` is the first group, so configuring or modifying a discrete attribute in the first certified group silently does nothing. Use an explicit undefined check.
```suggestion
    if (!attribute || attributeGroupIndex === undefined) return
```

src/static/locales/it/shared-components.json
```diff
@@ -339,12 +339,18 @@
...
     "idCopytooltipLabel": "ID copiato correttamente",
     "removeAttributeAriaLabel": "Rimuovi attributo {{attributeName}}",
     "thresholdLabel": "Chiamate API/giorno per fruitore",
-    "customizeBtn": "Personalizza soglia",
-    "changeBtn": "Modifica soglia",
+    "actions": {
+      "customizeThreshold": "Personalizza soglia",
+      "changeThreshold": "Modifica soglia",
+      "inspectAttributeDetails": "Dettagli attributo",
+      "changeCertifiedDiscreteAttribute": "Modifica attributo"
```
Alepazz: This key is `changeCertifiedDiscreteAttribute`, but the component reads `actions.modifyCertifiedDiscreteAttribute` (and the English file defines that name), so the Italian UI renders the raw key instead of the label. Align the key name:
```suggestion
      "modifyCertifiedDiscreteAttribute": "Modifica attributo"
```

Alepazz (changes requested):
Two small fixes before merge, see inline comments.

src/components/shared/AddAttributesToForm/AttributeGroup.tsx
DenisLantero: Yes, I think adding a test for `remapDescriptorAttributesToDescriptorAttributesSeed` that verifies `discreteConfig` is preserved would be great

DenisLantero (approved):
LGTM, just fix @Alepazz [comment](https://github.com/pagopa/pdnd-interop-frontend/pull/1935#discussion_r3386826526) before merging

src/components/shared/AddAttributesToForm/AttributeGroup.tsx
alten-dturus: I will add it in the PR where the other map functions are implemented and tested

Alepazz (approved):
LGTM

PR #1949, opened 2026-05-29
feat: initial risk analysis list commit (PIN-10187)

l-giacomini:
## Issue
- [PIN-10187](https://pagopa.atlassian.net/browse/PIN-10187)
- Added Risk Analysis list page
- Added Risk Analysis route
- Mocked SUBSCRIBE_RISK_ANALYSIS_INFO_COMPILE - SUBSCRIBE_RISK_ANALYSIS_COMPILE routes
- Added GET /purposes/riskAnalysis/assignments service
- Added tests

[PIN-10187]: https://pagopa.atlassian.net/browse/PIN-10187?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/shared/StatusChip.tsx
+++ b/src/components/shared/StatusChip.tsx
@@ -47,7 +47,7 @@ const CHIP_COLORS_PURPOSE: Record<PurposeVersionState, MUIColor> = {
 }
 
 const CHIP_COLORS_RISK_ANALYSIS: Record<
-  Exclude<RiskAnalysisSigningState, 'SIGNED' | 'REJECTED'>,
+  Exclude<RiskAnalysisSigningState, 'DRAFT' | 'SIGNED' | 'REJECTED'>,
   MUIColor
 > = {
   ASSIGNED: 'warning',
@@ -190,6 +190,11 @@ export const StatusChip: React.FC<StatusChipProps> = (props) => {
   if (props.for === 'purposeTemplate') {
     color = chipColors['purposeTemplate'][props.state]
     label = t(`status.purposeTemplate.${props.state}`)
+  }
+
+  if (props.for === 'riskAnalysis' && (props.state === 'ASSIGNED' || props.state === 'SUBMITTED')) {
+    color = chipColors['riskAnalysis'][props.state]
+    label = t(`status.riskAnalysis.${props.state}`)
   }
 
   return (
--- a/src/components/shared/StatusChip.tsx
+++ b/src/components/shared/StatusChip.tsx
@@ -15,6 +15,7 @@ import type {
   Purpose,
   PurposeTemplateState,
   PurposeVersionState,
+  RiskAnalysisSigningState,
 } from '@/api/api.generatedTypes'
 
 const CHIP_COLORS_E_SERVICE: Record<EServiceDescriptorState, MUIColor> = {
@@ -45,6 +46,14 @@ const CHIP_COLORS_PURPOSE: Record<PurposeVersionState, MUIColor> = {
   REJECTED: 'error',
 }
 
+const CHIP_COLORS_RISK_ANALYSIS: Record<
+  Exclude<RiskAnalysisSigningState, 'SIGNED' | 'REJECTED'>,
+  MUIColor
+> = {
+  ASSIGNED: 'warning',
+  SUBMITTED: 'info',
+}
+
 const CHIP_COLORS_DELEGATION: Record<DelegationState, MUIColor> = {
   ACTIVE: 'success',
   REJECTED: 'error',
@@ -73,6 +82,7 @@ const chipColors = {
   delegation: CHIP_COLORS_DELEGATION,
   eserviceTemplate: CHIP_COLORS_E_SERVICE_TEMPLATE,
   purposeTemplate: CHIP_COLORS_PURPOSE_TEMPLATE,
+  riskAnalysis: CHIP_COLORS_RISK_ANALYSIS,
 } as const
 
 type StatusChipProps = Omit<ChipProps, 'color' | 'label'> &
@@ -101,6 +111,10 @@ type StatusChipProps = Omit<ChipProps, 'color' | 'label'> &
     | {
         for: 'purposeTemplate'
         state: PurposeTemplateState
+      }
+    | {
+        for: 'riskAnalysis'
+        state: RiskAnalysisSigningState
       }
   )
 
--- a/src/components/shared/StatusChip.tsx
+++ b/src/components/shared/StatusChip.tsx
@@ -114,7 +114,7 @@ type StatusChipProps = Omit<ChipProps, 'color' | 'label'> &
       }
     | {
         for: 'riskAnalysis'
-        state: RiskAnalysisSigningState
+        state: Extract<RiskAnalysisSigningState, 'ASSIGNED' | 'SUBMITTED'>
       }
   )
 
--- a/src/pages/RiskAnalysisListPage/RiskAnalysisList.page.tsx
+++ b/src/pages/RiskAnalysisListPage/RiskAnalysisList.page.tsx
@@ -8,7 +8,7 @@ import {
   useAutocompleteTextInput,
 } from '@pagopa/interop-fe-commons'
 import { useQuery, keepPreviousData } from '@tanstack/react-query'
-import type { GetRiskAnalysisAssignmentsParams } from '@/api/api.generatedTypes'
+import type { GetRiskAnalysisAssignmentsParams, Purpose } from '@/api/api.generatedTypes'
 import { RiskAnalysisTable, RiskAnalysisTableSkeleton } from './components/RiskAnalysisTable'
 import { PurposeQueries } from '@/api/purpose'
 import { EServiceQueries } from '@/api/eservice'
@@ -56,7 +56,7 @@ const RiskAnalysisListPage: React.FC = () => {
     usePagination()
   const queryParams = { ...paginationParams, ...filtersParams }
 
-  const { data } = useQuery({
+  const { data, isFetching } = useQuery({
     ...PurposeQueries.getRiskAnalysisAssignments(queryParams),
     placeholderData: keepPreviousData,
   })
@@ -93,11 +93,11 @@ const RiskAnalysisListPage: React.FC = () => {
       ) : (
         <>
           <Filters {...filtersHandlers} />
-          <RiskAnalysisTableWrapper params={queryParams} />
+          <RiskAnalysisTableWrapper purposes={data?.results ?? []} isFetching={isFetching} />
           <Pagination
             {...paginationProps}
             rowPerPageOptions={rowPerPageOptions}
-            totalPages={getTotalPageCount(data?.pagination.totalCount)}
+            totalPages={getTotalPageCount(data?.pagination.totalCount ?? 0)}
           />
         </>
       )}
@@ -105,14 +105,13 @@ const RiskAnalysisListPage: React.FC = () => {
   )
 }
 
-const RiskAnalysisTableWrapper: React.FC<{ params: GetRiskAnalysisAssignmentsParams }> = ({
-  params,
+const RiskAnalysisTableWrapper: React.FC<{ purposes: Purpose[]; isFetching: boolean }> = ({
+  purposes,
+  isFetching,
 }) => {
-  const { data, isFetching } = useQuery(PurposeQueries.getRiskAnalysisAssignments(params))
-  if (!data && isFetching) return <RiskAnalysisTableSkeleton />
-  return (
-    <RiskAnalysisTable purposes={data?.results ?? []} data-testid="risk-analysis-table-component" />
-  )
+  if (isFetching && !purposes.length) return <RiskAnalysisTableSkeleton />
+
+  return <RiskAnalysisTable purposes={purposes} />
 }
 
 export default RiskAnalysisListPage
--- a/src/pages/RiskAnalysisListPage/RiskAnalysisList.page.tsx
+++ b/src/pages/RiskAnalysisListPage/RiskAnalysisList.page.tsx
@@ -61,19 +61,14 @@ const RiskAnalysisListPage: React.FC = () => {
     placeholderData: keepPreviousData,
   })
 
+  const hasActiveFilters =
+    (queryParams.eservicesIds?.length ?? 0) > 0 || Boolean(queryParams.signingState)
+
+  const isInitialEmptyState = (data?.results.length ?? 0) === 0 && !hasActiveFilters
+
   return (
     <PageContainer title={t('title')} description={t('description')}>
-      {data?.results.length !== 0 ? (
-        <>
-          <Filters {...filtersHandlers} />
-          <RiskAnalysisTableWrapper params={queryParams} />
-          <Pagination
-            {...paginationProps}
-            rowPerPageOptions={rowPerPageOptions}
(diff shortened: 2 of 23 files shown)


src/pages/RiskAnalysisListPage/RiskAnalysisList.page.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+  const { data } = useQuery({
+    ...PurposeQueries.getRiskAnalysisAssignments(queryParams),
+    placeholderData: keepPreviousData,
+  })
+
+  return (
+    <PageContainer title={t('title')} description={t('description')}>
+      {data?.results.length !== 0 ? (
+        <>
```
GorlemZ: This will mount `Filters` and `Pagination`  only when we have results: when a filter empties the results, the UI of the filters will disappear and the user will be blocked, without having the possibility to reset the filter. I'd delegate the empty state to the shared component `Table`

src/pages/RiskAnalysisListPage/RiskAnalysisList.page.tsx
```diff
@@ -0,0 +1,113 @@
...
+            sx={{
+              backgroundColor: 'white',
+              justifyContent: 'center',
+              borderRadius: 1,
+              display: 'flex',
+              py: 2,
+            }}
+          >
+            <Typography variant="body2" textAlign="center">
+              {tPurpose('noData.label')}
```
GorlemZ, line 89: This message is correct only if the user really has nothing assigned. Otherwhise, for example with a filter, the correct message should be "the search returned no results" or something similar imho

src/pages/RiskAnalysisListPage/components/RiskAnalysisTableRow.tsx
```diff
@@ -0,0 +1,96 @@
...
+    sentDate.getMonth() === today.getMonth() &&
+    sentDate.getFullYear() === today.getFullYear()
+
+  const formattedDate = sentDate
+    ? sentDate.toLocaleDateString(undefined, {
+        day: '2-digit',
+        month: '2-digit',
+        year: 'numeric',
+      })
+    : ''
```
GorlemZ: On a non-it browser the format will diverge, existing utils in the codebase use `formatDateString` (please check).
Also, `isToday` logic should be moved in `format.utils.ts`

src/pages/RiskAnalysisListPage/RiskAnalysisList.page.tsx
```diff
@@ -0,0 +1,113 @@
...
+  ])
+
+  const { paginationParams, paginationProps, getTotalPageCount, rowPerPageOptions } =
+    usePagination()
+  const queryParams = { ...paginationParams, ...filtersParams }
+
+  const { data } = useQuery({
+    ...PurposeQueries.getRiskAnalysisAssignments(queryParams),
+    placeholderData: keepPreviousData,
+  })
```
GorlemZ: The page and the wrapper launch the same query (in this file at line 106). React will dedupe for queryKey, but they are two subscriptions for the entire playload. Seems like this is useful only forr `data?.pagination.totalCount` after the ternary operator removal I've already commented, you can use somthing like:

```suggestion
  const { data: totalPages = 0 } = useQuery({
    ...PurposeQueries.getRiskAnalysisAssignments(queryParams),
    placeholderData: keepPreviousData,
    select: ({ pagination }) => getTotalPageCount(pagination.totalCount),
  })
```

src/pages/RiskAnalysisListPage/RiskAnalysisList.page.tsx
```diff
@@ -0,0 +1,113 @@
...
+import type { GetRiskAnalysisAssignmentsParams } from '@/api/api.generatedTypes'
+import { RiskAnalysisTable, RiskAnalysisTableSkeleton } from './components/RiskAnalysisTable'
+import { PurposeQueries } from '@/api/purpose'
+import { EServiceQueries } from '@/api/eservice'
+import { Box, Typography } from '@mui/material'
+
+const RiskAnalysisListPage: React.FC = () => {
+  const { t } = useTranslation('pages', { keyPrefix: 'riskAnalysisList' })
+  const { t: tPurpose } = useTranslation('purpose', { keyPrefix: 'riskAnalysisList' })
+
```
GorlemZ, line 20: why the i18n keys are splitted in two different namespaces?

src/pages/RiskAnalysisListPage/components/RiskAnalysisTableRow.tsx
```diff
@@ -0,0 +1,96 @@
...
+          </Typography>
+        ),
+        purpose.eservice.name,
+        purpose.eservice.producer.name,
+        purpose.reviewerWorkflow?.signingState ? (
+          <StatusChip
+            key={purpose.id}
+            for="riskAnalysis"
+            state={purpose.reviewerWorkflow.signingState}
+          />
```
GorlemZ, line 47: the key param can be removed here

GorlemZ (changes requested):

src/pages/RiskAnalysisListPage/RiskAnalysisList.page.tsx
l-giacomini: @GorlemZ I can't entirely delegate the empty state to the Table component, since the Figma (https://www.figma.com/design/CpRV3kPvFEWLXGtJUgWeZW/Interop-%E2%80%94-Delivery-FE---QA?node-id=7093-111926&t=jQEhqZQ1mlJLQz0U-4) requires a custom label which completely replaces the table.
This is true only if the initial data (without filters) is absent.
You are right about the second part: the filters would have made the user stuck on that label, it should be fixed now with this commit: https://github.com/pagopa/pdnd-interop-frontend/pull/1949/changes/5ed2bc5c7ab9d8fc8a8d84981fc978a4ae5f45c1

src/pages/RiskAnalysisListPage/RiskAnalysisList.page.tsx
l-giacomini: Fixed in: https://github.com/pagopa/pdnd-interop-frontend/pull/1949/changes/5ed2bc5c7ab9d8fc8a8d84981fc978a4ae5f45c1

src/pages/RiskAnalysisListPage/components/RiskAnalysisTableRow.tsx
l-giacomini: @GorlemZ formatDateString changes the Date with this pattern:

> const dateFormatter = new Intl.DateTimeFormat('it', {
>   day: '2-digit',
>   month: 'long',
>   year: 'numeric',
> })

resulting in a date like this: _03 giugno 2026_, which is not what the Figma requires (https://www.figma.com/design/CpRV3kPvFEWLXGtJUgWeZW/Interop-%E2%80%94-Delivery-FE---QA?node-id=7093-111831&t=jQEhqZQ1mlJLQz0U-4).

I've extracted this helper function into format.utils.ts, in this commit: https://github.com/pagopa/pdnd-interop-frontend/pull/1949/changes/eee695c574e9833b9f8ba927eb9995eed427838b

src/pages/RiskAnalysisListPage/RiskAnalysisList.page.tsx
l-giacomini: Fixed in: https://github.com/pagopa/pdnd-interop-frontend/pull/1949/changes/40a71d62f9bb0ac42acb5225f6b3fe57806fec74

src/pages/RiskAnalysisListPage/components/RiskAnalysisTableRow.tsx
l-giacomini: Fixed in: https://github.com/pagopa/pdnd-interop-frontend/pull/1949/changes/1e6d33fc00a611308f13bcb44bedddf1570e84fa

src/pages/RiskAnalysisListPage/RiskAnalysisList.page.tsx
l-giacomini: @GorlemZ I can see that "riskAnalysisList" from "pages" being a common pattern used throughout the application when a PageContainer component is instantiated, it just specifies the page title and description.
The other translation coming from "purpose" references to the feature itself.

src/pages/RiskAnalysisListPage/RiskAnalysisList.page.tsx
```diff
@@ -0,0 +1,117 @@
...
+  const { data, isFetching } = useQuery({
+    ...PurposeQueries.getRiskAnalysisAssignments(queryParams),
+    placeholderData: keepPreviousData,
+  })
+
+  const hasActiveFilters =
+    (queryParams.eservicesIds?.length ?? 0) > 0 || Boolean(queryParams.signingState)
+
+  const isInitialEmptyState = (data?.results.length ?? 0) === 0 && !hasActiveFilters
```
GorlemZ:
```suggestion
  const isInitialEmptyState = !!data && data.results.length === 0 && !hasActiveFilters
```

GorlemZ: This is worth testing: a simple test that mounts the page with `isFetching: true` / `data: undefined` and the assertion that **should't** expect the label `noData.label`. This test would fail now, try TDD! :D

src/pages/RiskAnalysisListPage/components/RiskAnalysisTableRow.tsx
```diff
@@ -0,0 +1,85 @@
...
+    : null
+
+  const formattedDate = sentDate
+    ? sentDate.toLocaleDateString(undefined, {
+        day: '2-digit',
+        month: '2-digit',
+        year: 'numeric',
+      })
+    : ''
+
```
GorlemZ, line 28: This is still not resolved: the formatting itself is still inline with `toLocaleDateString(undefined, …)`, which follows the browser locale; on a non-IT browser the day/month order flips (MM/dd vs dd/MM). Could we either pin the locale to 'it' or move a small `formatShortDate` helper into `format.utils.ts` next to `isToday`?

src/components/shared/StatusChip.tsx
```diff
@@ -102,6 +112,10 @@ type StatusChipProps = Omit<ChipProps, 'color' | 'label'> &
         for: 'purposeTemplate'
         state: PurposeTemplateState
       }
+    | {
+        for: 'riskAnalysis'
+        state: RiskAnalysisSigningState
+      }
```
GorlemZ, line 118: Nit / not blocking: this accepts the full `RiskAnalysisSigningState` but only ASSIGNED/SUBMITTED are handled below (other states would render an empty chip with no compile-time warning). If the assignments endpoint can only ever return those two, narrowing to state: `'ASSIGNED' | 'SUBMITTED'` would make it explicit (heads-up: the table row passes the full union, so it'd need a small guard). Fine to leave as-is if we're confident about the BE contract.

GorlemZ (changes requested):

src/components/shared/StatusChip.tsx
borgesis95: I Agree with this: If we handle only this two states use only them:

`Pick<RiskAnalysisSigningState, "ASSIGNED" | "SUBMITTED">`

src/pages/RiskAnalysisListPage/components/RiskAnalysisTable.tsx
```diff
@@ -0,0 +1,54 @@
+import { Table } from '@pagopa/interop-fe-commons'
+
+import { useTranslation } from 'react-i18next'
+import type { Purpose } from '@/api/api.generatedTypes'
+import { RiskAnalysisTableRow, RiskAnalysisTableRowSkeleton } from './RiskAnalysisTableRow'
+
+type RiskAnalysisTableProps = {
+  purposes: Array<Purpose>
```
borgesis95, line 8: suggestion: Could be an idea call this var `riskAnalysis` instead of `purposes`? Watching FIGMA we always find "analisi del rischio"

src/pages/RiskAnalysisListPage/components/RiskAnalysisTableRow.tsx
```diff
@@ -0,0 +1,85 @@
...
+  return (
+    <TableRow
+      cellData={[
+        isToday(sentDate) ? (
+          <Stack direction="row" spacing={1.5} alignItems="center">
+            <NotificationBadgeDot />
```
borgesis95: <NotificationBadgeDot />  should be shown if there is a notification not if `assignmentDate` is today. FIGMA is misleading in that case 
````suggestion
```

src/pages/RiskAnalysisListPage/RiskAnalysisList.page.tsx
l-giacomini: Test added in: https://github.com/pagopa/pdnd-interop-frontend/pull/1949/changes/5311a7d8c50385f51f1382da57178be5bb30c296

src/utils/format.utils.ts
```diff
@@ -26,3 +26,16 @@ export function minutesToSeconds(minutes: number) {
 export function secondsToMinutes(seconds: number) {
   return seconds / 60
 }
+
+export function isToday(date: Date | null) {
```
borgesis95: Can we use this https://date-fns.org/v4.4.0/docs/isToday? I believe that we have date-fns on our repository

src/pages/RiskAnalysisListPage/components/RiskAnalysisTableRow.tsx
l-giacomini: Fixed in: https://github.com/pagopa/pdnd-interop-frontend/pull/1949/changes/2a746473612db4aa5894363b651e09c1e109fac2

src/components/shared/StatusChip.tsx
l-giacomini: Fixed in: https://github.com/pagopa/pdnd-interop-frontend/pull/1949/changes/b0d40d211323affb9ee9ae482967ec2859062af0

src/pages/RiskAnalysisListPage/components/RiskAnalysisTable.tsx
l-giacomini: Sincerely I don't think it's worth it, the API actually returns an array of Purposes, changing this to RiskAnalysis would just make more confusion than benefits imho.

src/pages/RiskAnalysisListPage/components/RiskAnalysisTableRow.tsx
l-giacomini: Fixed in: https://github.com/pagopa/pdnd-interop-frontend/pull/1949/changes/49097b580a24489083568ac178efedc56f18ea4f

src/utils/format.utils.ts
l-giacomini: Fixed in: https://github.com/pagopa/pdnd-interop-frontend/pull/1949/changes/49097b580a24489083568ac178efedc56f18ea4f

GorlemZ (approved):
Approved! Please make sure you reach the min test coverage :)

borgesis95 (approved):

PR #1950, opened 2026-05-29
feat: risk analysis compile info page (PIN-10189)

l-giacomini:
## Issue
- [PIN-10189](https://pagopa.atlassian.net/browse/PIN-10189)
- Added Risk Analysis Compile Info page
- Added Risk Analysis mocked routes
- Added tests

[PIN-10189]: https://pagopa.atlassian.net/browse/PIN-10189?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/pages/RiskAnalysisInfoCompilePage/RiskAnalysisInfoCompile.page.tsx
+++ b/src/pages/RiskAnalysisInfoCompilePage/RiskAnalysisInfoCompile.page.tsx
@@ -1,0 +1,134 @@
+import { PurposeQueries } from '@/api/purpose'
+import { PageContainer, SectionContainer } from '@/components/layout/containers'
+import { Link, useNavigate, useParams } from '@/router'
+import React from 'react'
+import { useTranslation } from 'react-i18next'
+import { Button, Grid, Skeleton, Stack } from '@mui/material'
+import { useQuery } from '@tanstack/react-query'
+import { InformationContainer } from '@pagopa/interop-fe-commons'
+
+const RiskAnalysisInfoCompilePage: React.FC = () => {
+  const { t } = useTranslation('purpose', { keyPrefix: 'riskAnalysisInfoCompile' })
+  const { purposeId } = useParams<'PROVIDE_PURPOSE_DETAILS'>()
+  const navigate = useNavigate()
+
+  const { data: purpose, isLoading } = useQuery(PurposeQueries.getSingle(purposeId))
+
+  const handleBeginCompile = () => {
+    if (purpose?.id) {
+      navigate('SUBSCRIBE_RISK_ANALYSIS_COMPILE', {
+        params: { purposeId: purpose.id },
+      })
+    }
+  }
+
+  return (
+    <PageContainer
+      title={t('title')}
+      isLoading={isLoading}
+      backToAction={{
+        label: t('backToListBtn'),
+        to: 'SUBSCRIBE_RISK_ANALYSIS_LIST',
+      }}
+    >
+      <Grid container>
+        <Grid item xs={12}>
+          {!purpose ? (
+            <ProviderPurposeDetailsPageSkeleton />
+          ) : (
+            <Stack spacing={2}>
+              <SectionContainer title={t('generalInfoSection.label')}>
+                <Stack spacing={3}>
+                  <InformationContainer
+                    label={t('generalInfoSection.eService.label')}
+                    content={
+                      <Link
+                        to="SUBSCRIBE_CATALOG_VIEW"
+                        params={{
+                          eserviceId: purpose.eservice.id,
+                          descriptorId: purpose.eservice.descriptor.id,
+                        }}
+                        target="_blank"
+                      >
+                        {purpose.eservice.name}
+                      </Link>
+                    }
+                  />
+                  <InformationContainer
+                    label={t('generalInfoSection.producer.label')}
+                    content={purpose.eservice.producer.name}
+                  />
+                  <InformationContainer
+                    label={t('generalInfoSection.purposeName.label')}
+                    content={purpose.title}
+                  />
+                  <InformationContainer
+                    label={t('generalInfoSection.purposeDescription.label')}
+                    content={purpose.description}
+                  />
+                  <InformationContainer
+                    label={t('generalInfoSection.isFreeOfCharge.label')}
+                    content={
+                      purpose.isFreeOfCharge
+                        ? t('generalInfoSection.isFreeOfCharge.options.YES')
+                        : t('generalInfoSection.isFreeOfCharge.options.NO')
+                    }
+                  />
+                  <InformationContainer
+                    label={t('generalInfoSection.freeOfChargeReason.label')}
+                    content={purpose.freeOfChargeReason || ''}
+                  />
+                </Stack>
+              </SectionContainer>
+              <SectionContainer title={t('loadEstimationSection.label')}>
+                <Stack spacing={3}>
+                  <InformationContainer
+                    label={t('loadEstimationSection.dailyCalls.label')}
+                    content={`${purpose.currentVersion?.dailyCalls}`}
+                  />
+                </Stack>
+              </SectionContainer>
+            </Stack>
+          )}
+        </Grid>
+      </Grid>
+      <Stack direction="row" sx={{ mt: 4, justifyContent: 'right' }}>
+        <Button onClick={handleBeginCompile} variant="contained" type="button">
+          {t('beginCompileBtn')}
+        </Button>
+      </Stack>
+    </PageContainer>
+  )
+}
+
+const RiskAnalysisGeneralInfoSectionSkeleton: React.FC = () => (
+  <SectionContainer title="">
+    <Stack spacing={3}>
+      <Skeleton variant="text" width="40%" height={32} />
+      <Skeleton variant="rectangular" height={56} />
+      <Skeleton variant="rectangular" height={56} />
+      <Skeleton variant="rectangular" height={56} />
+      <Skeleton variant="rectangular" height={56} />
+    </Stack>
+  </SectionContainer>
+)
+
+const RiskAnalysisLoadEstimateSectionSkeleton: React.FC = () => (
+  <SectionContainer title="">
+    <Stack spacing={3}>
+      <Skeleton variant="text" width="30%" height={32} />
+      <Skeleton variant="rectangular" height={56} />
+    </Stack>
+  </SectionContainer>
+)
+
+const ProviderPurposeDetailsPageSkeleton: React.FC = () => {
+  return (
+    <Stack spacing={3}>
+      <RiskAnalysisGeneralInfoSectionSkeleton />
+      <RiskAnalysisLoadEstimateSectionSkeleton />
+    </Stack>
+  )
+}
+
+export default RiskAnalysisInfoCompilePage
--- a/src/pages/RiskAnalysisInfoCompilePage/RiskAnalysisInfoCompile.page.tsx
+++ b/src/pages/RiskAnalysisInfoCompilePage/RiskAnalysisInfoCompile.page.tsx
@@ -9,7 +9,7 @@ import { InformationContainer } from '@pagopa/interop-fe-commons'
 
 const RiskAnalysisInfoCompilePage: React.FC = () => {
   const { t } = useTranslation('purpose', { keyPrefix: 'riskAnalysisInfoCompile' })
-  const { purposeId } = useParams<'PROVIDE_PURPOSE_DETAILS'>()
+  const { purposeId } = useParams<'SUBSCRIBE_RISK_ANALYSIS_INFO_COMPILE'>()
   const navigate = useNavigate()
 
   const { data: purpose, isLoading } = useQuery(PurposeQueries.getSingle(purposeId))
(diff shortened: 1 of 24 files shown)


src/pages/RiskAnalysisInfoCompilePage/RiskAnalysisInfoCompile.page.tsx
```diff
@@ -0,0 +1,134 @@
...
+                </Stack>
+              </SectionContainer>
+              <SectionContainer title={t('loadEstimationSection.label')}>
+                <Stack spacing={3}>
+                  <InformationContainer
+                    label={t('loadEstimationSection.dailyCalls.label')}
+                    content={`${purpose.currentVersion?.dailyCalls}`}
+                  />
+                </Stack>
+              </SectionContainer>
```
DenisLantero, line 95: Aren't we missing the "Stima chiamate API" card here?

Here's the [figma reference](https://www.figma.com/design/CpRV3kPvFEWLXGtJUgWeZW/Interop-%E2%80%94-Delivery-FE---QA?node-id=7093-111528&t=l7Fq3T9vy6my8tNp-4)

src/pages/RiskAnalysisInfoCompilePage/__tests__/RiskAnalysisInfoCompile.page.test.tsx
```diff
@@ -0,0 +1,196 @@
...
+  it('should render page title', () => {
+    useQueryMock.mockReturnValue({
+      data: createMockPurpose(),
+      isLoading: false,
+    })
+
+    renderWithApplicationContext(<RiskAnalysisInfoCompilePage />, {
+      withReactQueryContext: true,
+      withRouterContext: true,
+    })
```
DenisLantero, line 50: The tests render the page component directly, so they do not cover route authorization, breadcrumbs, or back-button navigation. Jira explicitly asks for non-Valutatore blocking, back navigation, breadcrumb/title/cards/CTA rendering, and CTA navigation. Can we add at least one route-wrapper/AuthGuard test for the route role and one page test for back navigation/breadcrumb expectations?

GorlemZ (changes requested):
Two asks:
- Should the daily-calls value come from the waiting/draft version rather than currentVersion for these purposes? Worth confirming which version is the source of truth here.
- Could you add a test covering a purpose without currentVersion? It should fail as-is and guard whatever fix we land on.
Thanks!

src/pages/RiskAnalysisInfoCompilePage/RiskAnalysisInfoCompile.page.tsx
l-giacomini: @DenisLantero what do you mean? The "Stima chiamate API" card is exactly the piece of code that you've commented on.

This is the rendered result:

image

DenisLantero: You are right, my bad, I don't know how I didn't notice this 😂

src/pages/RiskAnalysisInfoCompilePage/__tests__/RiskAnalysisInfoCompile.page.test.tsx
l-giacomini: @DenisLantero Since we still don't have the roles available on the development server, I've decided (in agreement with @borgesis95) to write route guards and related tests after having merged the basic tasks in the feature branch.
I've left TODO comments on src/router/routes.tsx to keep this in mind.

l-giacomini:
> Two asks:
> 
> * Should the daily-calls value come from the waiting/draft version rather than currentVersion for these purposes? Worth confirming which version is the source of truth here.
> * Could you add a test covering a purpose without currentVersion? It should fail as-is and guard whatever fix we land on.
>   Thanks!

@GorlemZ for the first question: I'm not sure what do you mean with the waiting/draft version, the user in this section cannot modify the purpose, also I can only see apiCalls in the currentVersion object from the response.
For the second point, I've checked the dailyCalls field throughout the application's code, it is set to 1 which is the minimum if missing. 
I've added the test in: https://github.com/pagopa/pdnd-interop-frontend/pull/1950/changes/a20f0faabf9d356ffed14756f39b74c2ce5de587 and set it to 1 as a fallback value.

DenisLantero (approved):

src/pages/RiskAnalysisInfoCompilePage/RiskAnalysisInfoCompile.page.tsx
```diff
@@ -0,0 +1,134 @@
...
+import { Button, Grid, Skeleton, Stack } from '@mui/material'
+import { useQuery } from '@tanstack/react-query'
+import { InformationContainer } from '@pagopa/interop-fe-commons'
+
+const RiskAnalysisInfoCompilePage: React.FC = () => {
+  const { t } = useTranslation('purpose', { keyPrefix: 'riskAnalysisInfoCompile' })
+  const { purposeId } = useParams<'SUBSCRIBE_RISK_ANALYSIS_INFO_COMPILE'>()
+  const navigate = useNavigate()
+
+  const { data: purpose, isLoading } = useQuery(PurposeQueries.getSingle(purposeId))
```
GorlemZ: What happens if this query returns a 404 and `isLoading` is `false`? We are not throwing!
```suggestion
  const { data: purpose, isLoading } = useQuery({
  ...PurposeQueries.getSingle(purposeId),
  throwOnError: true,
  })
```

src/pages/RiskAnalysisInfoCompilePage/RiskAnalysisInfoCompile.page.tsx
```diff
@@ -0,0 +1,134 @@
...
+              </SectionContainer>
+            </Stack>
+          )}
+        </Grid>
+      </Grid>
+      <Stack direction="row" sx={{ mt: 5, justifyContent: 'right' }}>
+        <Button onClick={handleBeginCompile} variant="contained" type="button">
+          {t('beginCompileBtn')}
+        </Button>
+      </Stack>
```
GorlemZ: This button is outside the loading conditional block, so it is rendered and clickable while the page is still showing the skeleton. Could we disable it until the purpose is loaded?

src/pages/RiskAnalysisInfoCompilePage/RiskAnalysisInfoCompile.page.tsx
```diff
@@ -0,0 +1,134 @@
...
+                    label={t('generalInfoSection.isFreeOfCharge.label')}
+                    content={
+                      purpose.isFreeOfCharge
+                        ? t('generalInfoSection.isFreeOfCharge.options.YES')
+                        : t('generalInfoSection.isFreeOfCharge.options.NO')
+                    }
+                  />
+                  <InformationContainer
+                    label={t('generalInfoSection.freeOfChargeReason.label')}
+                    content={purpose.freeOfChargeReason || ''}
```
GorlemZ: When the purpose is **not** free (`isFreeOfCharge === false`) there's no `freeOfChargeReason`, so this renders an empty "Free of charge reason" row. Could we render it only when relevant?

src/pages/RiskAnalysisInfoCompilePage/RiskAnalysisInfoCompile.page.tsx
```diff
@@ -0,0 +1,134 @@
...
+const RiskAnalysisLoadEstimateSectionSkeleton: React.FC = () => (
+  <SectionContainer title="">
+    <Stack spacing={3}>
+      <Skeleton variant="text" width="30%" height={32} />
+      <Skeleton variant="rectangular" height={56} />
+    </Stack>
+  </SectionContainer>
+)
+
+const ProviderPurposeDetailsPageSkeleton: React.FC = () => {
```
GorlemZ: Are we sure about this naming?

GorlemZ (changes requested):

GorlemZ:
> > Two asks:
> > 
> > * Should the daily-calls value come from the waiting/draft version rather than currentVersion for these purposes? Worth confirming which version is the source of truth here.
> > * Could you add a test covering a purpose without currentVersion? It should fail as-is and guard whatever fix we land on.
> >   Thanks!
> 
> @GorlemZ for the first question: I'm not sure what do you mean with the waiting/draft version, the user in this section cannot modify the purpose, also I can only see apiCalls in the currentVersion object from the response. For the second point, I've checked the dailyCalls field throughout the application's code, it is set to 1 which is the minimum if missing. I've added the test in: [a20f0fa](https://github.com/pagopa/pdnd-interop-frontend/commit/a20f0faabf9d356ffed14756f39b74c2ce5de587) and set it to 1 as a fallback value.

Thanks for digging in! On the "waiting/draft version" point: `currentVersion` isn't empty because the data is missing, it's how the BFF computes it. `getCurrentVersion` excludes `WAITING_FOR_APPROVAL`/`REJECTED` versions (`backend-for-frontend/.../purposeService.ts:96`), and the purposes that reach this page are exactly the ones still under review, so their `currentVersion` is `undefined`. But the daily-calls they entered are still in the response under  `waitingForApprovalVersion` (the BFF maps it, same shape, so it has `dailyCalls` too).
So `?? 1` shows a hardcoded "1" for every purpose in approval instead of the real estimate. Could we fall back to the pending version? Something like:

```
content={`${purpose.currentVersion?.dailyCalls ?? purpose.waitingForApprovalVersion?.dailyCalls ?? '-'}`}
```

(1 is the right creation minimum, but here it's a read-only display of an existing value.)

src/pages/RiskAnalysisInfoCompilePage/RiskAnalysisInfoCompile.page.tsx
borgesis95: Good catch! @l-giacomini could be ask to designers what should render in this case?

borgesis95 (approved):
LGTM: fix comments that @GorlemZ raised before to merge

src/pages/RiskAnalysisInfoCompilePage/RiskAnalysisInfoCompile.page.tsx
l-giacomini: Fixed in: https://github.com/pagopa/pdnd-interop-frontend/pull/1950/changes/359ce97012b76b68991e12ea30b46a6e96695aa2

src/pages/RiskAnalysisInfoCompilePage/RiskAnalysisInfoCompile.page.tsx
l-giacomini: Fixed in: https://github.com/pagopa/pdnd-interop-frontend/pull/1950/changes/359ce97012b76b68991e12ea30b46a6e96695aa2

src/pages/RiskAnalysisInfoCompilePage/RiskAnalysisInfoCompile.page.tsx
l-giacomini: Fixed in: https://github.com/pagopa/pdnd-interop-frontend/pull/1950/changes/359ce97012b76b68991e12ea30b46a6e96695aa2

src/pages/RiskAnalysisInfoCompilePage/RiskAnalysisInfoCompile.page.tsx
l-giacomini: Comment added: https://www.figma.com/design/CpRV3kPvFEWLXGtJUgWeZW?node-id=7093-111500#1789063367

l-giacomini:
> > > Two asks:
> > > 
> > > * Should the daily-calls value come from the waiting/draft version rather than currentVersion for these purposes? Worth confirming which version is the source of truth here.
> > > * Could you add a test covering a purpose without currentVersion? It should fail as-is and guard whatever fix we land on.
> > >   Thanks!
> > 
> > 
> > @GorlemZ for the first question: I'm not sure what do you mean with the waiting/draft version, the user in this section cannot modify the purpose, also I can only see apiCalls in the currentVersion object from the response. For the second point, I've checked the dailyCalls field throughout the application's code, it is set to 1 which is the minimum if missing. I've added the test in: [a20f0fa](https://github.com/pagopa/pdnd-interop-frontend/commit/a20f0faabf9d356ffed14756f39b74c2ce5de587) and set it to 1 as a fallback value.
> 
> Thanks for digging in! On the "waiting/draft version" point: `currentVersion` isn't empty because the data is missing, it's how the BFF computes it. `getCurrentVersion` excludes `WAITING_FOR_APPROVAL`/`REJECTED` versions (`backend-for-frontend/.../purposeService.ts:96`), and the purposes that reach this page are exactly the ones still under review, so their `currentVersion` is `undefined`. But the daily-calls they entered are still in the response under `waitingForApprovalVersion` (the BFF maps it, same shape, so it has `dailyCalls` too). So `?? 1` shows a hardcoded "1" for every purpose in approval instead of the real estimate. Could we fall back to the pending version? Something like:
> 
> ```
> content={`${purpose.currentVersion?.dailyCalls ?? purpose.waitingForApprovalVersion?.dailyCalls ?? '-'}`}
> ```
> 
> (1 is the right creation minimum, but here it's a read-only display of an existing value.)

@GorlemZ  I've added the suggested fallback in: https://github.com/pagopa/pdnd-interop-frontend/pull/1950/changes/032cf9ae223ee14c241c536cfc53acca1c165262

GorlemZ (approved):

src/pages/RiskAnalysisInfoCompilePage/RiskAnalysisInfoCompile.page.tsx
l-giacomini: The design team have decided to hide the field if isFreeOfCharge is set to false, logic added in: https://github.com/pagopa/pdnd-interop-frontend/pull/1950/changes/905353dcf6d24c5255abbca2bbefa40e480980a2

PR #1951, opened 2026-05-29
feature: implement new logic for checking if attributes discrete are compliant (PIN-10219)

alten-dturus:
## 🔗 Issue

[PIN-10219](https://pagopa.atlassian.net/browse/PIN-10219)

## 📝 Description / Context

This PR introduced the new logic for certified discrete attributes to see the attribute is compliant and so owned. The other changes are to adapt the attributes and attributes groups check utilis functions to include the certified discrete attributes.

## 🛠 List of changes

- Changed attributes.utils to add certified discrete attributes logic

[PIN-10219]: https://pagopa.atlassian.net/browse/PIN-10219?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


borgesis95 (approved):

src/utils/attribute.utils.ts
```diff
@@ -116,22 +119,45 @@ export function isAttributeOwned(
...
     case 'certified':
-      return !isAttributeRevoked(
-        'certified',
-        match as CertifiedTenantAttribute | CertifiedDiscreteTenantAttribute
+      return (
+        !isAttributeRevoked(
+          'certified',
+          match as CertifiedTenantAttribute | CertifiedDiscreteTenantAttribute
+        ) &&
+        (match.kind === 'CERTIFIED_DISCRETE' && options?.discreteConfig
```
DenisLantero: I think this fallback makes the new discrete check inconsistent in the UI. A `CERTIFIED_DISCRETE` tenant attribute is treated as owned whenever `discreteConfig` is missing, and several row-level checks still call `isAttributeOwned('certified', attribute.id, ownedCertifiedAttributes)` without passing the descriptor config. The group can be red because the threshold fails, while the individual attribute still renders checked.

Could we pass `attribute.discreteConfig` in the certified row checks too, or make this helper return false for `CERTIFIED_DISCRETE` when the descriptor config is missing? Examples affected:
https://github.com/pagopa/pdnd-interop-frontend/blob/feature/PIN-10219_checking-logic-for-discrete-attributes/src/components/shared/ReadOnlyDescriptorAttributes.tsx#L212
https://github.com/pagopa/pdnd-interop-frontend/blob/feature/PIN-10219_checking-logic-for-discrete-attributes/src/pages/ConsumerAgreementDetailsPage/components/ConsumerAgreementDetailsAttributesSectionsList/ConsumerAgreementDetailsCertifiedAttributesSection.tsx#L64

src/utils/__tests__/attribute.utils.test.ts
```diff
@@ -104,6 +128,46 @@ describe('attribute utils', () => {
...
+      })
+      const result = isAttributeOwned('certified', 'attribute-id', [attributeMock], {
+        discreteConfig: { comparator: 'GT', threshold: 50 },
+      })
+      expect(result).toBe(false)
+    })
+
+    it('should not be considered owned if the attribute is in the owned attribute array but it is not compliant to discrete config (CERTIFIED_DISCRETE)', () => {
+      const attributeMock = createCertifiedDiscreteTenantAttribute({
+        revocationTimestamp: 'timestamp',
```
DenisLantero: This test says it covers the "not compliant to discrete config" case, but the fixture is also revoked, so it would stay green even if `isAttributeCompliantWithDiscreteConfig` were never called. Could we set `revocationTimestamp: undefined` here so the false result depends only on the discrete comparator?

Alepazz (commented):
Agree with the open point about discrete certified attributes rendering unchecked at row level while the group is green: threading the descriptor discreteConfig into the per-row certified checks would fix it. One more nit: isAttributeCompliantWithDiscreteConfig only has tests for GT and LT, the other comparators (GTE, LTE, EQ, NE) are new behavior worth covering.

DenisLantero (commented):
I think the row-level certified checks are still missing the descriptor `discreteConfig`.

After the latest change, `isAttributeOwned('certified', ...)` returns `false` for `CERTIFIED_DISCRETE` when no config is passed:
https://github.com/pagopa/pdnd-interop-frontend/blob/086b023ab90adf31488a38227ee516c55c0522a0/src/utils/attribute.utils.ts#L134-L139

But the row-level checks still call it without passing `attribute.discreteConfig`, so a compliant discrete attribute can render unchecked while its group is green:
https://github.com/pagopa/pdnd-interop-frontend/blob/086b023ab90adf31488a38227ee516c55c0522a0/src/components/shared/ReadOnlyDescriptorAttributes.tsx#L275-L278
https://github.com/pagopa/pdnd-interop-frontend/blob/086b023ab90adf31488a38227ee516c55c0522a0/src/pages/ConsumerAgreementDetailsPage/components/ConsumerAgreementDetailsAttributesSectionsList/ConsumerAgreementDetailsCertifiedAttributesSection.tsx#L61-L65
https://github.com/pagopa/pdnd-interop-frontend/blob/086b023ab90adf31488a38227ee516c55c0522a0/src/pages/ProviderAgreementDetailsPage/components/ProviderAgreementDetailsAttributesSectionsList/ProviderAgreementDetailsCertifiedAttributesSection.tsx#L62-L65

Could we pass the descriptor `discreteConfig` into these certified row checks too?

alten-dturus:
> I think the row-level certified checks are still missing the descriptor `discreteConfig`.
> 
> After the latest change, `isAttributeOwned('certified', ...)` returns `false` for `CERTIFIED_DISCRETE` when no config is passed:
> 
> https://github.com/pagopa/pdnd-interop-frontend/blob/086b023ab90adf31488a38227ee516c55c0522a0/src/utils/attribute.utils.ts#L134-L139
> 
> But the row-level checks still call it without passing `attribute.discreteConfig`, so a compliant discrete attribute can render unchecked while its group is green:
> 
> https://github.com/pagopa/pdnd-interop-frontend/blob/086b023ab90adf31488a38227ee516c55c0522a0/src/components/shared/ReadOnlyDescriptorAttributes.tsx#L275-L278
> 
> 
> https://github.com/pagopa/pdnd-interop-frontend/blob/086b023ab90adf31488a38227ee516c55c0522a0/src/pages/ConsumerAgreementDetailsPage/components/ConsumerAgreementDetailsAttributesSectionsList/ConsumerAgreementDetailsCertifiedAttributesSection.tsx#L61-L65
> 
> 
> https://github.com/pagopa/pdnd-interop-frontend/blob/086b023ab90adf31488a38227ee516c55c0522a0/src/pages/ProviderAgreementDetailsPage/components/ProviderAgreementDetailsAttributesSectionsList/ProviderAgreementDetailsCertifiedAttributesSection.tsx#L62-L65
> 
> Could we pass the descriptor `discreteConfig` into these certified row checks too?

Ok, i understand now. I added that discreteConfig. The only problem is in ReadOnlyDescriptorAttribute because outside the api called to retrieve the attributes is not changed to return Certified and CertifiedDiscrete but only certified. I will talk to BE

src/utils/attribute.utils.ts
```diff
@@ -95,43 +89,81 @@ export function isAttributeOwned(
...
+            !isAttributeRevoked(
+              'certified',
+              attributeMatched as CertifiedDiscreteTenantAttribute
+            ) &&
+            (options?.discreteConfig
+              ? isAttributeCompliantWithDiscreteConfig(
+                  attributeMatched as CertifiedDiscreteTenantAttribute,
+                  options.discreteConfig
+                )
+              : false)
```
Alepazz, line 133: This `: false` branch (an owned, non revoked `CERTIFIED_DISCRETE` attribute called without a `discreteConfig`) is the core invariant the row-level certified checks rely on, but it has no unit test: every `isAttributeOwned` discrete test passes a `discreteConfig`. Could we add a case asserting that an owned, non revoked `CERTIFIED_DISCRETE` returns `false` when no `discreteConfig` is provided? It would lock the invariant and also help recover the new-code coverage that is currently failing the quality gate.

src/utils/attribute.utils.ts
```diff
@@ -95,43 +89,81 @@ export function isAttributeOwned(
...
    * we check if the attribute is expired.
    */
-
   switch (kind) {
     case 'certified':
-      return !isAttributeRevoked(
-        'certified',
-        match as CertifiedTenantAttribute | CertifiedDiscreteTenantAttribute
-      )
+      const isOwned = match(attributeMatched.kind)
```
Alepazz: `const isOwned` is declared directly inside the `case 'certified':` clause without a block, so it lives in the whole `switch` scope (the `no-case-declarations` smell) and is an unnecessary temporary. Could we return the `match(...)` expression directly, or wrap the case body in `{ }`?

Alepazz (approved):
LGTM

DenisLantero (approved):
LGTM, maybe just address what @Alepazz said

PR #1956, opened 2026-06-03
bugfix: agreement consumer validation (PIN-10234)

DenisLantero:
## 🔗 Issue

[PIN-10234](https://pagopa.atlassian.net/browse/PIN-10234)

## 📝 Description / Context

Fixes the agreement consumer selection dialog shown when requesting e-service consumption via delegation. When the selected party is cleared, the dialog now shows the required field validation instead of the missing certified attributes warning.

## 🛠 List of changes

- Show the missing certified attributes warning only when a consumer is actually selected.
- Keep the create action submittable when the consumer field is empty so form validation can display the required field error.
- Add a regression test for clearing the consumer autocomplete in the create agreement dialog.

## 🧪 How to test

- Open the e-service catalog and request consumption for an e-service with an active consumer delegation.
- In the “Create draft consumption request” dialog, clear the selected party from the dropdown.
- Click “Create draft” and verify that the required field error is shown, without the certified attributes warning.
- Select a party without the required certified attributes and verify that the warning is still shown.

[PIN-10234]: https://pagopa.atlassian.net/browse/PIN-10234?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/dialogs/DialogSelectAgreementConsumer/__tests__/DialogSelectAgreementConsumer.test.tsx
+++ b/src/components/dialogs/DialogSelectAgreementConsumer/__tests__/DialogSelectAgreementConsumer.test.tsx
@@ -1,0 +1,67 @@
+import { rest } from 'msw'
+import { setupServer } from 'msw/node'
+import userEvent from '@testing-library/user-event'
+import { screen, waitFor } from '@testing-library/react'
+import { DialogSelectAgreementConsumer } from '../DialogSelectAgreementConsumer'
+import { BACKEND_FOR_FRONTEND_URL } from '@/config/env'
+import { queryClient } from '@/config/query-client'
+import { mockUseJwt, renderWithApplicationContext } from '@/utils/testing.utils'
+
+const server = setupServer(
+  rest.get(`${BACKEND_FOR_FRONTEND_URL}/delegations`, (_, res, ctx) => {
+    return res(ctx.json({ results: [] }))
+  }),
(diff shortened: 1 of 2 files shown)


src/components/dialogs/DialogSelectAgreementConsumer/__tests__/DialogSelectAgreementConsumer.test.tsx
```diff
@@ -0,0 +1,68 @@
...
+afterEach(() => {
+  server.resetHandlers()
+  queryClient.clear()
+  vi.clearAllMocks()
+})
+
+afterAll(() => server.close())
+
+describe('DialogSelectAgreementConsumer', () => {
+  it('shows the required field validation instead of the missing attributes alert when the consumer is cleared', async () => {
```
borgesis95, line 36: is necessary test that attributes alert is not showed? it's good if we write this test only related to bug that QA opened but in my opinion it's not necessary

DenisLantero: This indeed comes from a QA bug, here's the [issue](https://pagopa.atlassian.net/browse/PIN-10234)

borgesis95 (approved):

PR #1958, opened 2026-06-03
refactor: add new mid type for attributes forms (PIN-10238)

alten-dturus:
## 🔗 Issue

[PIN-10238](https://pagopa.atlassian.net/browse/PIN-10238)

## 📝 Description / Context

This PR add two new intermediate types for attribute that are needed to manage the fact that `AttributeAutocomplete` uses `CompactAttribute` as options. The type `FormDescriptorAttribute` a mid type from `CompactAttribute` and `DescriptorAttribute` (and `DescriptorAttributeSeed`) while `FormDescriptorAttributes` is for `FormDescriptorAttribute` the same that `DescrptorAttributes` is for `DescriptorAttribute` so an object where the attribute are divided by the corresponding key.

## 🛠 List of changes

- Implemented new types
- Fix wrong type in `AttributeAutocomplete`
- Implemented new functions to map `DescriptorAttribute` to `FormDescriptorAttribute` and `FormDescriptorAttribute` to `DescriptorAttributeSeed`

[PIN-10238]: https://pagopa.atlassian.net/browse/PIN-10238?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


borgesis95 (changes requested):
Could you help me better to what I need to review here? https://github.com/pagopa/pdnd-interop-frontend/pull/1958#issue-4578963582 this and JIRA's task are lack of information

src/utils/attribute.utils.ts
```diff
@@ -241,6 +245,56 @@ export function hasAllDescriptorAttributes(
...
 }
 
+/**
+ * This should be temporary, it is here because from the BFF we get the attributes in a different
+ * format than the one we need to send to the API.
+ * @param descriptorAttributes - The attributes to remap.
+ * @returns The remapped attributes.
+ */
+// TODO
+export const remapFormDescriptorAttributesToDescriptorAttributesSeed = (
```
borgesis95: I'd like to remove "re" as prefix. This should be `mapFormDescriptorAttributesToDescriptorAttributesSeed` etc...

src/pages/ProviderEServiceCreatePage/components/EServiceCreateStepThresholds/EServiceCreateStepThresholds.tsx
```diff
@@ -169,7 +187,8 @@ export const EServiceCreateStepThresholds: React.FC<ActiveStepProps> = () => {
             dailyCallsTotal: values.dailyCallsTotal,
             agreementApprovalPolicy: descriptor.agreementApprovalPolicy,
             description: descriptor.description,
-            attributes: remapDescriptorAttributesToDescriptorAttributesSeed(attributes),
+            // attributes: remapDescriptorAttributesToDescriptorAttributesSeed(attributes), TODO
+            attributes: remapFormDescriptorAttributesToDescriptorAttributesSeed(attributes),
           },
           { onSuccess: forward }
         )
```
borgesis95, line 186: The equality check compares form-shaped attributes against descriptor-shaped attributes in `EServiceCreateStepThresholds.tsx:147` and` EServiceCreateStepThresholds.tsx:153` : `compareObjects` method is watching descriptor.attributes (with type `DescriptorAttributes`) and ```
```
const attributes: {
    certified: FormDescriptorAttribute[][];
    verified: FormDescriptorAttribute[][];
    declared: FormDescriptorAttribute[][];
}
```

is right?

src/pages/ProviderEServiceCreatePage/components/EServiceCreateStepThresholds/EServiceCreateStepThresholds.tsx
```diff
@@ -169,7 +187,8 @@ export const EServiceCreateStepThresholds: React.FC<ActiveStepProps> = () => {
             dailyCallsTotal: values.dailyCallsTotal,
             agreementApprovalPolicy: descriptor.agreementApprovalPolicy,
             description: descriptor.description,
-            attributes: remapDescriptorAttributesToDescriptorAttributesSeed(attributes),
+            // attributes: remapDescriptorAttributesToDescriptorAttributesSeed(attributes), TODO
```
borgesis95:
```suggestion
```

src/pages/ProviderEServiceCreatePage/components/EServiceCreateStepThresholds/EServiceCreateStepThresholds.tsx
```diff
@@ -130,14 +144,17 @@ export const EServiceCreateStepThresholds: React.FC<ActiveStepProps> = () => {
...
-    const newDescriptorData: ProducerEServiceDescriptor = {
-      ...descriptor,
+    const newDescriptorData = {
       dailyCallsPerConsumer: values.dailyCallsPerConsumer,
       dailyCallsTotal: values.dailyCallsTotal,
       attributes,
     }
 
-    const areDescriptorsEquals = compareObjects(newDescriptorData, descriptor)
+    const areDescriptorsEquals = compareObjects(newDescriptorData, {
```
Alepazz, line 147: `compareObjects` iterates the keys of the first argument and runs `isEqual` per key. Here `newDescriptorData.attributes` is `FormDescriptorAttributes` (no `description` / `explicitAttributeVerification`), while `descriptor.attributes` is the raw `DescriptorAttributes`, so `isEqual` is always false when the descriptor has attributes. As a result `areDescriptorsEquals` never becomes true and the update is sent even when nothing changed. Could you compare like for like, for example remapping the descriptor side with `remapDescriptorAttributesToFormDescriptorAttributes(descriptor.attributes)` (or comparing the seeds)?

src/components/shared/AddAttributesToForm/AttributeGroup.tsx
```diff
@@ -62,12 +65,12 @@ export const AttributeGroup: React.FC<AttributeGroupProps> = ({
...
 
-  const { watch, setValue, getValues } = useFormContext<{ attributes: DescriptorAttributes }>()
+  const { watch, setValue, getValues } = useFormContext<{ attributes: FormDescriptorAttributes }>()
   const attributeGroups = watch(`attributes.${attributeKey}`)
 
-  const handleAddAttributeToGroup = (attribute: DescriptorAttribute) => {
+  const handleAddAttributeToGroup = (attribute: CompactAttribute) => {
     const newAttributeGroups = [...attributeGroups]
-    newAttributeGroups[groupIndex].push(attribute)
+    newAttributeGroups[groupIndex].push(attribute as FormDescriptorAttribute)
```
Alepazz: `CompactAttribute` ({ id, name, kind }) is already structurally assignable to `FormDescriptorAttribute` (the extra fields are optional), so this cast can be dropped.
```suggestion
    newAttributeGroups[groupIndex].push(attribute)
```

Alepazz (commented):
A couple of comments inline, the main one is about the change detection that looks like it regresses.

src/pages/ProviderEServiceCreatePage/components/EServiceCreateStepThresholds/EServiceCreateStepThresholds.tsx
alten-dturus: That's an error, I think using mapDescriptorAttributesToFormDescriptorAttributes(descriptor.attributes) on the objects to compare solves that

src/pages/ProviderEServiceCreatePage/components/EServiceCreateStepThresholds/EServiceCreateStepThresholds.tsx
alten-dturus: Yes, you're right, I will use the mapping function

DenisLantero: Could we add a focused test for the “unchanged descriptor with attributes” case here? The current test at https://github.com/pagopa/pdnd-interop-frontend/blob/0520cf03fd0e16f2de47141d65d31163ce092716/src/pages/ProviderEServiceCreatePage/components/EServiceCreateStepThresholds/__tests__/EServiceCreateStepThresholds.test.tsx#L61 passes with `descriptor` undefined, so it would not catch the compare-shape regression that was discussed here. A descriptor with at least one attribute should submit, skip the API, and call `forward()`.

src/utils/attribute.utils.ts
```diff
@@ -241,6 +245,49 @@ export function hasAllDescriptorAttributes(
   return descriptorAttributes.every(isGroupFullfilled)
 }
 
+export const mapFormDescriptorAttributesToDescriptorAttributesSeed = (
```
DenisLantero, line 280: Could we add direct tests for the new mapping helpers? This is the core behavior of the PR, and the drawer test mocks `mapFormDescriptorAttributesToDescriptorAttributesSeed`, so regressions like dropping `discreteConfig` or losing the form shape would still pass. A small `attribute.utils.test.ts` case for descriptor -> form and form -> seed should be enough.

DenisLantero (approved):
LGTM, check inline comments

src/pages/ProviderEServiceCreatePage/components/EServiceCreateStepThresholds/EServiceCreateStepThresholds.tsx
alten-dturus: I changed the test adding a descriptor with a CERTIFIED attribute and the result doesn't change. Maybe I don't understand properly what do you mean. Please check the new changes if they are like you said

borgesis95 (approved):
LGTM, before to merge resolve conflicts please

src/pages/ProviderEServiceCreatePage/components/EServiceCreateStepThresholds/EServiceCreateStepThresholds.tsx
DenisLantero: Yes, this is the scenario I meant, now the descriptor is defined and has at least one attribute, so it covers the shape mismatch case.

Could you also add an assertion on `forward()`? The important behavior here is not only that we skip `updateVersionDraft`, but also that the unchanged descriptor follows the “nothing to save” path and moves to the next step. Without checking `forward()`, the test would still pass if the submit handler simply returned early without calling the API, which would be a broken user flow.

DenisLantero (approved):

Alepazz (approved):
My two points are addressed, thanks. Approving. Still worth resolving the test follow-up [Denis raised](https://github.com/pagopa/pdnd-interop-frontend/pull/1958#discussion_r3386870636): the unchanged-descriptor test should also assert that forward() is called, not just that the API call is skipped.

src/pages/ProviderEServiceCreatePage/components/EServiceCreateStepThresholds/EServiceCreateStepThresholds.tsx
alten-dturus: Check on forward added

PR #1960, opened 2026-06-03
Feature/pin 10193 approve risk analysis dialog

l-giacomini:
## Issue
- [PIN-10191](https://pagopa.atlassian.net/browse/PIN-10191)
- Added Risk Analysis Approve Dialog
- Added POST /purposes/{purposeId}/riskAnalysis/sign service
- Mocked routes
- Added tests

[PIN-10191]: https://pagopa.atlassian.net/browse/PIN-10191?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/dialogs/DialogApproveRiskAnalysis.tsx
+++ b/src/components/dialogs/DialogApproveRiskAnalysis.tsx
@@ -23,22 +23,30 @@ export const DialogApproveRiskAnalysis: React.FC<DialogApproveRiskAnalysisProps>
   })
   const { closeDialog } = useDialog()
   const navigate = useNavigate()
-  const { mutate: signRiskAnalysis } = PurposeMutations.useSignRiskAnalysis()
+  const { mutate: signRiskAnalysis, isPending } = PurposeMutations.useSignRiskAnalysis()
+
+  const handleCloseDialog = () => {
+    if (!isPending) {
+      closeDialog()
+    }
+  }
 
   const onProceed = () => {
     signRiskAnalysis(
       { purposeId },
       {
         onSuccess() {
-          navigate('SUBSCRIBE_RISK_ANALYSIS_APPROVAL_SUCCESS', { params: { purposeId } })
+          navigate('SUBSCRIBE_RISK_ANALYSIS_APPROVAL_SUCCESS', {
+            params: { purposeId },
+          })
+          handleCloseDialog()
         },
       }
     )
-    closeDialog()
   }
 
   return (
-    <Dialog aria-labelledby={ariaLabelId} open onClose={closeDialog} fullWidth>
+    <Dialog aria-labelledby={ariaLabelId} open onClose={handleCloseDialog} fullWidth>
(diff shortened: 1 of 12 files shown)


src/components/dialogs/DialogApproveRiskAnalysis.tsx
```diff
@@ -0,0 +1,62 @@
...
+      </DialogContent>
+
+      <DialogActions>
+        <Button type="button" variant="outlined" onClick={closeDialog}>
+          {t('actions.cancel')}
+        </Button>
+        <Button variant="contained" onClick={onProceed}>
+          {t('actions.confirm')}
+        </Button>
+      </DialogActions>
```
DenisLantero, line 67: These buttons are never disabled and CTA has no loading state, should we add it?

src/components/dialogs/DialogApproveRiskAnalysis.tsx
```diff
@@ -0,0 +1,62 @@
...
+    signRiskAnalysis(
+      { purposeId },
+      {
+        onSuccess() {
+          navigate('SUBSCRIBE_RISK_ANALYSIS_APPROVAL_SUCCESS', { params: { purposeId } })
+        },
+      }
+    )
+    closeDialog()
+  }
```
DenisLantero, line 46: In [PIN-10193](https://pagopa.atlassian.net/browse/PIN-10191) the expected behavior says that, on confirm, the dialog should stay open during the request, disable the CTAs to avoid double clicks, close only on success, and remain open on error showing the project error pattern.

Here we call `closeDialog()` immediately after `signRiskAnalysis(...)`, so the dialog closes before knowing whether the POST succeeded or failed. That means the error case cannot be shown in the dialog and the user cannot retry from there.

Can we move `closeDialog()` into the mutation `onSuccess` and use the mutation pending/error state to keep the dialog open, disable the actions during loading, and handle the failure case?

DenisLantero (changes requested):
A few blocking points, check the inline comments, maybe fix the SonarQube analysys as well

src/components/dialogs/DialogApproveRiskAnalysis.tsx
l-giacomini: Fixed in: https://github.com/pagopa/pdnd-interop-frontend/pull/1960/changes/f7e8093774b9443bd397b2ab7f58da7a6ce21136

src/components/dialogs/DialogApproveRiskAnalysis.tsx
l-giacomini: Fixed in: https://github.com/pagopa/pdnd-interop-frontend/pull/1960/changes/f7e8093774b9443bd397b2ab7f58da7a6ce21136

DenisLantero (approved):

borgesis95 (approved):

PR #1978, opened 2026-06-08
fix: minor manual archiving fixes (PIN-10257, PIN-10273, PIN-10275, PIN-10276)

Alepazz:
## 🔗 Issue

- [PIN-10257](https://pagopa.atlassian.net/browse/PIN-10257)
- [PIN-10273](https://pagopa.atlassian.net/browse/PIN-10273)
- [PIN-10275](https://pagopa.atlassian.net/browse/PIN-10275)
- [PIN-10276](https://pagopa.atlassian.net/browse/PIN-10276)

## 📝 Description / Context

Collection of minor, independent fixes that emerged from the manual archiving QA (children of PIN-9936). Small, low-risk changes, kept separate from the heavier header/menu action-logic changes (which go in a dedicated PR).

## 🛠 List of changes

- **PIN-10257** — Archiving date format unified to numeric (dd/mm/yyyy): badge tooltip (`ArchivingScheduleBadge`) and all status banners (`getEServiceDescriptorAlertSpec`), aligned with the dialogs that already used the numeric format.
- **PIN-10273** — Date in the "Archivia e-service" dialog aligned with the server value: uses `calculateArchivableOn` (like the "Archivia versione" dialog) instead of `addDays`, which produced a date off by one day.
- **PIN-10275** — The scheduled-archiving badge now disappears when the version is `ARCHIVED`: gated on the `ARCHIVING`/`ARCHIVING_SUSPENDED` states in the e-service detail pages (provider and consumer) and in the "Stato versioni" drawer. Introduced the shared helper `isDescriptorPendingArchiving` (also reused in `StatusChip` and the provider detail page to remove duplicated conditions).
- **PIN-10276** — EN copy of the "Fruition Request Draft Creation" dialog: replaced the (non-rendered) markdown bold `**...**` with `<strong>...</strong>`, consistent with the Italian.
- Tests: unit tests for the archiving-state helper and a test that the badge is not shown for `ARCHIVED` versions in the drawer.

## 🧪 How to test

- **Numeric dates**: with a version/e-service being archived, check that the badge tooltip and banners show the date as `dd/mm/yyyy`.
- **Archivia e-service dialog**: the notice end date matches the `archivableOn` returned by the server (same day as the "Archivia versione" dialog).
- **Post-archiving badge**: once a version is `ARCHIVED`, the "scheduled archiving" badge is no longer visible (page header + "Stato versioni" drawer); only the "archived on …" banner remains.
- **EN copy**: open the fruition request creation dialog in EN and check the e-service name/version in bold, without `**`.

[PIN-10257]: https://pagopa.atlassian.net/browse/PIN-10257
[PIN-10273]: https://pagopa.atlassian.net/browse/PIN-10273
[PIN-10275]: https://pagopa.atlassian.net/browse/PIN-10275
[PIN-10276]: https://pagopa.atlassian.net/browse/PIN-10276

### DIFF
--- a/src/components/dialogs/__tests__/DialogShowEserviceVersionsList.test.tsx
+++ b/src/components/dialogs/__tests__/DialogShowEserviceVersionsList.test.tsx
@@ -117,4 +117,19 @@ describe('DialogShowEserviceVersionsList', () => {
 
     expect(screen.queryByTestId('ArchiveIcon')).not.toBeInTheDocument()
   })
+
+  it('does not show the archive icon for ARCHIVED descriptors even if archivableOn is set', () => {
+    renderDialog({
+      descriptors: [
+        makeDescriptor({
+          id: 'd1',
+          version: '1',
+          state: 'ARCHIVED',
+          archivableOn: '2026-12-31',
+        }),
+      ],
+    })
+
+    expect(screen.queryByTestId('ArchiveIcon')).not.toBeInTheDocument()
+  })
 })
--- a/src/pages/ProviderEServiceDetailsPage/ProviderEServiceDetails.page.tsx
+++ b/src/pages/ProviderEServiceDetailsPage/ProviderEServiceDetails.page.tsx
@@ -14,7 +14,11 @@ import { NewPageContainer } from '@/components/layout/containers/NewPageContaine
 import { useDialog } from '@/stores'
 import { useDrawerState } from '@/hooks/useDrawerState'
 import { EServiceVersionSelectorDrawer } from '@/components/shared/EServiceVersionSelectorDrawer'
-import { getActiveDescriptor, getViewLatestVersionTargetId } from '@/utils/eservice.utils'
+import {
+  getActiveDescriptor,
+  getViewLatestVersionTargetId,
+  isDescriptorBeingArchived,
+} from '@/utils/eservice.utils'
 import { ProviderEServiceDetailsAlerts } from './components/ProviderEServiceDetailsTab/ProviderEServiceDetailsAlerts'
 
 const ProviderEServiceDetailsPage: React.FC = () => {
@@ -52,8 +56,7 @@ const ProviderEServiceDetailsPage: React.FC = () => {
   )
 
   const isActiveDescriptor = descriptor?.id === activeDescriptor?.id
-  const isEServiceBeingArchived =
-    activeDescriptor?.state === 'ARCHIVING' || activeDescriptor?.state === 'ARCHIVING_SUSPENDED'
+  const isEServiceBeingArchived = isDescriptorBeingArchived(activeDescriptor?.state)
 
   const hasMultipleVersions = (descriptor?.eservice.descriptors?.length ?? 0) > 1
 
@@ -113,7 +116,9 @@ const ProviderEServiceDetailsPage: React.FC = () => {
                 isActiveDescriptor,
               },
               archivingScheduleInfo:
-                descriptor.archivingSchedule?.archivableOn && descriptor.archivingSchedule?.scope
+                isDescriptorBeingArchived(descriptor.state) &&
+                descriptor.archivingSchedule?.archivableOn &&
+                descriptor.archivingSchedule?.scope
                   ? {
                       archivableOn: descriptor.archivingSchedule.archivableOn,
                       scope: descriptor.archivingSchedule.scope,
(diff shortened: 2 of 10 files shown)


src/components/dialogs/__tests__/DialogShowEserviceVersionsList.test.tsx
```diff
@@ -117,4 +117,19 @@ describe('DialogShowEserviceVersionsList', () => {
 
     expect(screen.queryByTestId('ArchiveIcon')).not.toBeInTheDocument()
   })
+
+  it('does not show the archive icon for ARCHIVED descriptors even if archivableOn is set', () => {
```
borgesis95, line 121: is this compliant with FIGMA?

src/utils/eservice.utils.ts
```diff
@@ -114,3 +114,7 @@ export function calculateArchivableOn(now: Date, gracePeriodDays: number): Date
   d.setUTCHours(0, 0, 0, 0)
   return d
 }
+
+export function isDescriptorBeingArchived(state: EServiceDescriptorState | undefined): boolean {
```
borgesis95: isDescriptorBeingArchived implies an active, ongoing process — which fits `ARCHIVING` well. However, `ARCHIVING_SUSPENDED` means the archiving is paused/on hold, so the descriptor is not actively being archived at that moment.

A more accurate name could be:

- `isDescriptorInArchivingProcess` — covers both the active and suspended phases of the archiving flow

src/utils/__tests__/eservice.utils.test.ts
```diff
@@ -146,6 +147,22 @@ describe('calculateArchivableOn utility function testing', () => {
   })
 })
 
+describe('isDescriptorBeingArchived utility function testing', () => {
+  it('returns true for ARCHIVING and ARCHIVING_SUSPENDED', () => {
+    expect(isDescriptorBeingArchived('ARCHIVING')).toBe(true)
+    expect(isDescriptorBeingArchived('ARCHIVING_SUSPENDED')).toBe(true)
+  })
+
+  it('returns false for other states and undefined', () => {
```
borgesis95:
```suggestion
  it.each([
'PUBLISHED',
'ARCHIVED',
'SUSPENDED',
'DEPRECATED',
'DRAFT',
undefined,
])('returns false for state %s', (state) => {
expect(isDescriptorBeingArchived(state)).toBe(false)
})
```

src/pages/ProviderEServiceDetailsPage/ProviderEServiceDetails.page.tsx
```diff
@@ -113,7 +116,9 @@ const ProviderEServiceDetailsPage: React.FC = () => {
                 isActiveDescriptor,
               },
               archivingScheduleInfo:
-                descriptor.archivingSchedule?.archivableOn && descriptor.archivingSchedule?.scope
+                isDescriptorBeingArchived(descriptor.state) &&
```
borgesis95: Could you use isEServiceBeingArchived that you declared before instead of  `isDescriptorBeingArchived(descriptor.state)`? Should be equal

src/components/dialogs/__tests__/DialogShowEserviceVersionsList.test.tsx
Alepazz: Yes, it is related to [this comment on Slack](https://pagopaspa.slack.com/archives/C0AMNUSH5MG/p1780666330562489).
I checked with the be team and they stated that `archivableOn` is returned even if the e-service is ARCHIVED, in order to keep a history of the archiving operations

src/utils/eservice.utils.ts
Alepazz: Good point on the naming. Since both `ARCHIVING` and `ARCHIVING_SUSPENDED` represent a scheduled-but-not-yet-completed archiving (they are exactly the states that carry an `archivingSchedule`, as opposed to the final `ARCHIVED`), what about `isDescriptorPendingArchiving`? It covers both the active and the paused phase, and reads a bit more concise than `isDescriptorInArchivingProcess`. Happy to rename if you agree.

src/pages/ProviderEServiceDetailsPage/ProviderEServiceDetails.page.tsx
Alepazz: These are not equivalent: `isEServiceBeingArchived` is derived from `activeDescriptor.state`, while here we need the state of the descriptor currently being viewed (same as the adjacent `statusChip`/version/`archivingSchedule`, which all read from `descriptor`). They diverge when viewing a non-active version, e.g. a single version under DESCRIPTOR-scope archiving while the active one is PUBLISHED, so I would keep `isDescriptorBeingArchived(descriptor.state)` here.

src/utils/eservice.utils.ts
borgesis95: Agree with `isDescriptorPendingArchiving` 👍

borgesis95 (approved):
Before merging update method's name as we discussed

PR #1983, opened 2026-06-08
fix: show success toast when canceling e-service archiving (PIN-10294)

Alepazz:
## 🔗 Issue

[PIN-10294](https://pagopa.atlassian.net/browse/PIN-10294)

## 📝 Description / Context

When canceling the archiving of an e-service (the "Annulla archiviazione e-service" dialog), Figma specifies a success toast with the text "L'e-service non è più in fase di archiviazione", but the frontend did not show it: the `useCancelEserviceArchiving` mutation did not define `meta.successToastLabel`, and the global handler shows the success toast only if that label is present. This PR adds the missing toast for that specific action.

## 🛠 List of changes

- `useCancelEserviceArchiving` (`src/api/eservice/eservice.mutations.ts`): added `meta.successToastLabel`.
- New i18n key `eservice.cancelEserviceArchiving.outcome.success` in `mutations-feedback.json` (IT: "L'e-service non è più in fase di archiviazione"; EN: "The e-service is no longer being archived").

## 🧪 How to test

- Put an e-service into archiving (e-service scope).
- From the provider detail, open "Annulla archiviazione e-service" and confirm.
- Check that the success toast appears: "L'e-service non è più in fase di archiviazione" (EN: "The e-service is no longer being archived").

[PIN-10294]: https://pagopa.atlassian.net/browse/PIN-10294


src/api/eservice/eservice.mutations.ts
```diff
@@ -213,8 +213,14 @@ function useScheduleArchiveEservice() {
 }
 
 function useCancelEserviceArchiving() {
+  const { t } = useTranslation('mutations-feedback', {
+    keyPrefix: 'eservice.cancelEserviceArchiving',
+  })
   return useMutation({
     mutationFn: EServiceServices.cancelEserviceArchiving,
+    meta: {
```
borgesis95, line 235: What about error?

Alepazz: The error copy isn't defined in Figma, so there's no text specified for an error toast in this flow. On top of that, `develop` already has 4 archiving mutations without an error toast (the schedule/cancel ones for both descriptor and e-service), so this isn't specific to this change. If you think it's worth adding, I can check with the Design team for the error copy.

borgesis95 (approved):

PR #2001, opened 2026-06-11
feat: show reviewer full name in purpose approval dialog

LucaPierfederici:
## 🔗 Issue
[PIN-10301](https://pagopa.atlassian.net/browse/PIN-10301)

## 📝 Description / Context
The "Request approval" dialog (Step 3, option 2 — `ADMIN_WRITES_REVIEWER_SIGNS`) was showing the raw reviewer id instead of the reviewer's full name. The BE now exposes the reviewer info inline on the reviewer workflow (`ReviewerWorkflow.reviewers: CompactUser[]`), so we can display the reviewer's name and surname, aligning this dialog with the rest of the purpose create/edit flow (assignment step, options, compilation dialog), which already used name + surname.

## 🛠 List of changes
- `DialogRequestPurposeApproval` now receives the `reviewer` (`CompactUser`) instead of `reviewerId`, and builds the full name internally.
- `DialogRequestPurposeApprovalProps`: `reviewerId: string` → `reviewer: CompactUser`.
- `PurposeEditStepRiskAnalysis` reads the reviewer directly from `reviewerWorkflow.reviewers[0]`, dropping the `reviewerIds[0]` lookup; logs and no-ops if the reviewer info is missing (BE contract violation).
- i18n (`it`/`en`): approval dialog description placeholder `{{reviewerId}}` → `{{reviewerName}}`.
- Updated unit tests for the dialog and the risk-analysis step accordingly.

## 🧪 How to test
- Create/edit a purpose as admin and, in Step 2, choose option 2 ("Compilazione in autonomia e approvazione del valutatore"), selecting a reviewer.
- Go to Step 3, fill in the risk analysis, and click "Richiedi approvazione".
- Verify the dialog body shows the reviewer's **name and surname** (not the id).
- Confirm the action and verify the save + submit + redirect to the purpose summary still works.

[PIN-10301]: https://pagopa.atlassian.net/browse/PIN-10301

### DIFF
--- a/src/pages/ConsumerPurposeEditPage/components/PurposeEditStepRiskAnalysis/PurposeEditStepRiskAnalysis.tsx
+++ b/src/pages/ConsumerPurposeEditPage/components/PurposeEditStepRiskAnalysis/PurposeEditStepRiskAnalysis.tsx
@@ -137,21 +137,20 @@ export const PurposeEditStepRiskAnalysis: React.FC<ActiveStepProps> = ({ back })
   }
 
   const handleRequestApproval = (answers: Record<string, string[]>) => {
-    const reviewerId = purpose.reviewerWorkflow?.reviewerIds?.[0]
-    if (!reviewerId) {
-      // BE contract: option 2 always assigns at least one reviewer
-      // (RiskAnalysisAssignmentSeed.reviewerIds has @minItems 1). If we land
-      // here the purpose is malformed; log loudly and no-op rather than
-      // crashing the route — this is a UI action handler, not a place to
-      // throw to the ErrorBoundary.
+    const reviewer = purpose.reviewerWorkflow?.reviewers?.[0]
+    if (!reviewer) {
+      // BE contract: option 2 always assigns at least one reviewer.
+      // If we land here the purpose is malformed;
+      // log loudly and no-op rather than crashing the route —
+      // this is a UI action handler, not a place to throw to the ErrorBoundary.
       console.error(
-        'PurposeEditStepRiskAnalysis: reviewerIds is missing on a purpose in ADMIN_WRITES_REVIEWER_SIGNS mode'
+        'PurposeEditStepRiskAnalysis: reviewers is missing on a purpose in ADMIN_WRITES_REVIEWER_SIGNS mode'
       )
       return
     }
     openDialog({
       type: 'requestPurposeApproval',
-      reviewerId,
+      reviewer,
       // Chain must live in the parent: putting it in the dialog would race
       // closeDialog(), and the second mutate() would silently no-op against a
       // destroyed observer.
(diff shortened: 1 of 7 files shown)


src/pages/ConsumerPurposeEditPage/components/PurposeEditStepRiskAnalysis/PurposeEditStepRiskAnalysis.tsx
```diff
@@ -137,21 +137,20 @@ export const PurposeEditStepRiskAnalysis: React.FC<ActiveStepProps> = ({ back })
...
 
   const handleRequestApproval = (answers: Record<string, string[]>) => {
-    const reviewerId = purpose.reviewerWorkflow?.reviewerIds?.[0]
-    if (!reviewerId) {
-      // BE contract: option 2 always assigns at least one reviewer
-      // (RiskAnalysisAssignmentSeed.reviewerIds has @minItems 1). If we land
-      // here the purpose is malformed; log loudly and no-op rather than
-      // crashing the route — this is a UI action handler, not a place to
-      // throw to the ErrorBoundary.
+    const reviewer = purpose.reviewerWorkflow?.reviewers?.[0]
```
Alepazz, line 140: The guard now keys off `reviewerWorkflow.reviewers`, which the generated contract marks optional (`reviewers?: CompactUser[]`), whereas the previous code keyed off the required `reviewerIds`. If the BE ever returns a valid option-2 workflow with `reviewerIds` populated but the embedded `reviewers` array absent, the "Request approval" button would silently no-op (log + return) where it previously opened the dialog. Could you confirm the BE always populates `reviewers` whenever `reviewerIds` is set on the reviewer workflow? If that invariant holds, this guard is fine as a defensive check; if not, the approval flow needs a fallback (for instance resolving the name from the tenant reviewers list, as the assignment step does).

Alepazz (commented):
Looks good, the reviewer name now matches the rest of the flow. One question on the BE contract before merging, see inline.

src/pages/ConsumerPurposeEditPage/components/PurposeEditStepRiskAnalysis/PurposeEditStepRiskAnalysis.tsx
LucaPierfederici: You are right, at the contract level the invariant doesn't hold (reviewers optional, reviewerIds required).

But I checked the BFF (`enrichPurposeReviewerWorkflow` in `purposeService.ts`): reviewers is enriched with one entry per reviewerId on the consumer side, and only omitted when the request is not consumer-side (producer/delegator). 
This dialog is reachable only from `SUBSCRIBE_PURPOSE_EDIT` (consumer admin or delegate), which is always consumer-side, so reviewers is guaranteed to be populated here.

So I'd keep the guard as a defensive check but no fallback is needed for this flow. 
WDYT?

borgesis95 (approved):

Alepazz (approved):
Clear, thanks for checking the BFF side. All good for me, resolving.

PR #2002, opened 2026-06-11
feat: update notification banner (PIN-10324)

DenisLantero:
## 🔗 Issue

[PIN-10324](https://pagopa.atlassian.net/browse/PIN-10324)

## 📝 Description / Context

This PR replaces the existing notifications banner with the new product updates banner for e-service archiving and new user roles.

## 🛠 List of changes

- Renamed the banner component, hook, test, and locale namespace from notifications-specific naming to `ProductUpdatesBanner`.
- Updated the banner copy in Italian and English.
- Replaced the notification configuration action with two external documentation links.
- Added dedicated documentation link constants for e-service archiving and user roles.
- Added localized accessible labels for the external-link actions.
- Updated the banner test to cover close behavior and both external CTA actions by accessible name.

## 🧪 How to test

- Start the app and make sure the product updates banner is visible, clearing `notificationsBannerDismissedUntil` from local storage if needed.
- Open the consumer e-service catalog page and verify that the banner shows the new PDND archiving/new roles copy.
- Click both banner actions and verify they open the expected documentation pages in a new tab.
- Close the banner and verify it is dismissed with the existing behavior.

[PIN-10324]: https://pagopa.atlassian.net/browse/PIN-10324?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/hooks/bannerHooks/useProductUpdatesBanner.ts
```diff
@@ -5,8 +5,8 @@ import { useTranslation } from 'react-i18next'
 
 const STORAGE_KEY = 'notificationsBannerDismissedUntil'
```
Alepazz: This new banner keeps the previous banner's dismissal identity: the same `notificationsBannerDismissedUntil` storage key, `bannerKey: 'notification'`, and the same `getNotificationsBannerConfigJson()` config (`notifications-window/data.json`). Since `useBaseBanner` hides the banner when the stored value equals the remote `endString`, any user who already dismissed the old notifications banner under the same window will not see this new product-updates announcement. Could you confirm the plan here? Either the remote `notifications-window` config is republished with a new window (new `end`, so the prior dismissal no longer matches), or this banner should use a dedicated storage key (for instance `productUpdatesBannerDismissedUntil`) to force a one-time re-display and track its own dismissal.

src/components/shared/__tests__/ProductUpdatesBanner.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+describe('Checks product updates banner alert', () => {
+  beforeEach(() => {
+    vi.clearAllMocks()
+  })
+
+  it('renders product updates banner with message and closes on button click', async () => {
+    const user = userEvent.setup()
+    const windowOpenSpy = vi.spyOn(window, 'open').mockImplementation(() => null)
+
+    vi.spyOn(useProductUpdatesBannerModule, 'useProductUpdatesBanner').mockReturnValue(
```
Alepazz, line 30: This test fully mocks `useProductUpdatesBanner`, so the new hook wiring (translation prefix, storage key, banner key, remote config query, priority) is not exercised. This is also the behavior most at risk from reusing the previous banner's storage key and config, so a small test asserting the hook passes the expected `storageKey` and config to `useBaseBanner` would lock it down. Not blocking on its own, but worth adding while the hook is being touched.

Alepazz (commented):
Looks good overall, the rename and the copy/accessibility updates are clean. Two points to confirm before merging, see inline.

src/hooks/bannerHooks/useProductUpdatesBanner.ts
DenisLantero: Good point, I didn't notice this and we indeed should use a new key, fixed in https://github.com/pagopa/pdnd-interop-frontend/commit/a3727697.

src/components/shared/__tests__/ProductUpdatesBanner.test.tsx
DenisLantero: Fixed here https://github.com/pagopa/pdnd-interop-frontend/commit/a3727697.

borgesis95 (approved):

Alepazz (approved):

PR #2006, opened 2026-06-12
performance: add lazy loading for attribute details drawer (PIN-10336)

alten-dturus:
## 🔗 Issue

[PIN-10336](https://pagopa.atlassian.net/browse/PIN-10336)

## 📝 Description / Context

This PR include the lazy loading of attribute for AttributeDetailsDrawer. The prefetch query starts on menu actions icon hover, then the query is inside the drawer. It is implemented a skeleton inside the drawer to show if the data are not ready to display.

## 🛠 List of changes

- Add prefetch query on menu icon hover
- Add skeleton in AttributeDetailsDrawer

[PIN-10336]: https://pagopa.atlassian.net/browse/PIN-10336?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/layout/containers/AttributeContainer.tsx
+++ b/src/components/layout/containers/AttributeContainer.tsx
@@ -13,9 +13,9 @@ import { ButtonNaked } from '@pagopa/mui-italia'
 import CheckCircleIcon from '@mui/icons-material/CheckCircle'
 import RemoveCircleOutlineIcon from '@mui/icons-material/RemoveCircleOutline'
 import { AttributeQueries } from '@/api/attribute'
-import { InformationContainer } from '@pagopa/interop-fe-commons'
+import { InformationContainer, InformationContainerSkeleton } from '@pagopa/interop-fe-commons'
 import { useTranslation } from 'react-i18next'
-import { useQuery } from '@tanstack/react-query'
+import { useQuery, useQueryClient } from '@tanstack/react-query'
 import { FEATURE_FLAG_ATTRIBUTE_CERTIFIED_DISCRETE } from '@/config/env'
 import { ActionMenu } from '@/components/shared/ActionMenu'
 import type { ActionItemButton } from '@/types/common.types'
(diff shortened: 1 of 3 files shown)


src/components/layout/containers/AttributeContainer.tsx
```diff
@@ -199,11 +209,11 @@ export const AttributeContainer = <
           )}
         </Card>
       </Stack>
-      {!isLoadingCompleteAttribute && completeAttribute && (
+      {alreadyPrefetched.current && (
```
DenisLantero, line 212: Non blocking, but I think this can break the inspect action because opening the drawer depends on prefetch having already happened.

For example, if the user activates “inspect attribute details” without `handlePrefetchAttribute` running first, the menu action sets `isDetailsDrawerOpen` to `true`, but the drawer is still not mounted because `alreadyPrefetched.current` is `false`.

https://github.com/pagopa/pdnd-interop-frontend/blob/fe3de977f631bd1b0ed617688a31c9813cadf492/src/components/layout/containers/AttributeContainer.tsx#L108-L110

In practice, the code says “open the drawer”, but this gate says “don’t even create the drawer unless prefetch already happened”. Could we make the click path self-contained by calling `handlePrefetchAttribute()` before `setIsDetailsDrawerOpen(true)`?

DenisLantero (approved):
LGTM, check inline comment

src/components/layout/containers/AttributeContainer.tsx
alten-dturus: But the `handlePrefetchAttribute` is called when we `focus` or `hover` on the `menuActions icon` so we cannot click on openDetails without calling the prefetch because it is before everytime or am I wrong?

DenisLantero: Yes that's true, that's why this is a non blocking comment, my point was more about making it more robust, but the fix is not really necessary.

borgesis95 (approved):

PR #2007, opened 2026-06-15
fix: remove save draft call (PIN-10342)

LucaPierfederici:
## 🔗 Issue
[PIN-10342](https://pagopa.atlassian.net/browse/PIN-10342)

## 📝 Description / Context
In the purpose creation form, "Compilazione in autonomia e approvazione da valutatore" flow, sending the purpose for approval used to fire two requests: a save-draft (`updatePurpose`) followed by the submit-for-approval (`submitRiskAnalysis`). The backend has unified both into the second call, so the save-draft request is now redundant and is removed.

## 🛠 List of changes
- Removed the superfluous `saveDraft`/`updatePurpose` call from the "Richiedi approvazione" handler in `PurposeEditStepRiskAnalysis`; `submitRiskAnalysis` is now invoked directly with `onSuccess: goToSummary`.

## 🧪 How to test
- Open a purpose in the "Compilazione in autonomia e approvazione da valutatore" (reviewer-approval) draft flow and reach the risk analysis step.
- Click "Richiedi approvazione" and confirm in the dialog.
- In the network tab, verify a single request is fired (only the risk-analysis submit, no preceding save-draft) and that on success you are redirected to the summary page.

[PIN-10342]: https://pagopa.atlassian.net/browse/PIN-10342

### DIFF
--- a/src/pages/ConsumerPurposeEditPage/components/PurposeEditStepRiskAnalysis/PurposeEditStepRiskAnalysis.tsx
+++ b/src/pages/ConsumerPurposeEditPage/components/PurposeEditStepRiskAnalysis/PurposeEditStepRiskAnalysis.tsx
@@ -156,17 +156,13 @@ export const PurposeEditStepRiskAnalysis: React.FC<ActiveStepProps> = ({ back })
       // closeDialog(), and the second mutate() would silently no-op against a
       // destroyed observer.
       onConfirm: () => {
-        saveDraft(answers, {
-          onSuccess: () => {
-            submitRiskAnalysis(
-              {
-                purposeId: purpose.id,
-                riskAnalysisForm: { version: riskAnalysis.version, answers },
-              },
-              { onSuccess: goToSummary }
-            )
+        submitRiskAnalysis(
+          {
+            purposeId: purpose.id,
+            riskAnalysisForm: { version: riskAnalysis.version, answers },
           },
-        })
+          { onSuccess: goToSummary }
+        )
       },
     })
   }
--- a/src/pages/ConsumerPurposeEditPage/components/PurposeEditStepRiskAnalysis/__tests__/PurposeEditStepRiskAnalysis.test.tsx
+++ b/src/pages/ConsumerPurposeEditPage/components/PurposeEditStepRiskAnalysis/__tests__/PurposeEditStepRiskAnalysis.test.tsx
@@ -300,7 +300,7 @@ describe('PurposeEditStepRiskAnalysis', () => {
     })
   })
 
-  it('in option 2 opens the requestPurposeApproval dialog and the dialog onConfirm runs the save+submit+navigate chain', () => {
+  it('in option 2 opens the requestPurposeApproval dialog and the dialog onConfirm runs the submit+navigate chain', () => {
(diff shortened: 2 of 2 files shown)


borgesis95 (approved):

src/pages/ConsumerPurposeEditPage/components/PurposeEditStepRiskAnalysis/PurposeEditStepRiskAnalysis.tsx
```diff
@@ -156,17 +156,13 @@ export const PurposeEditStepRiskAnalysis: React.FC<ActiveStepProps> = ({ back })
       // closeDialog(), and the second mutate() would silently no-op against a
```
Alepazz, line 156: This rationale comment (the three lines above `onConfirm`) is now stale: it describes a two-call chain and warns that "the second mutate() would silently no-op", but after dropping the save-draft call `onConfirm` fires a single `submitRiskAnalysis`, so there is no chain and no second mutate anymore. Since the project avoids comments in TS/TSX, simplest to remove the whole block.

Alepazz (approved):
LGTM. Left one inline suggestion below.

src/pages/ConsumerPurposeEditPage/components/PurposeEditStepRiskAnalysis/PurposeEditStepRiskAnalysis.tsx
```diff
@@ -156,17 +156,13 @@ export const PurposeEditStepRiskAnalysis: React.FC<ActiveStepProps> = ({ back })
...
-        saveDraft(answers, {
-          onSuccess: () => {
-            submitRiskAnalysis(
-              {
-                purposeId: purpose.id,
-                riskAnalysisForm: { version: riskAnalysis.version, answers },
-              },
-              { onSuccess: goToSummary }
-            )
+        submitRiskAnalysis(
```
DenisLantero, line 159: This direct submit path still uses the existing `submitRiskAnalysis` mutation copy, but that mutation documents and tells the user that the risk analysis was already saved before submit:
https://github.com/pagopa/pdnd-interop-frontend/blob/b15a1fab384cf004cbb55965e4d9f74876c7938d/src/api/purpose/purpose.mutations.ts#L237-L250
https://github.com/pagopa/pdnd-interop-frontend/blob/b15a1fab384cf004cbb55965e4d9f74876c7938d/src/static/locales/it/mutations-feedback.json#L639-L643
https://github.com/pagopa/pdnd-interop-frontend/blob/b15a1fab384cf004cbb55965e4d9f74876c7938d/src/static/locales/en/mutations-feedback.json#L639-L643

Now that the save happens inside the unified submit request, if that request fails the toast can no longer claim that the analysis was saved. Could we update the mutation comment and the IT/EN error copy accordingly?

LucaPierfederici: solved in: https://github.com/pagopa/pdnd-interop-frontend/commit/f7cc6b8f513e8bdc11c92c529df5c857d981af11

PR #2009, opened 2026-06-15
bugfix: fix typography bugs inside drawers (PIN-10346)

alten-dturus:
## 🔗 Issue

[PIN-10346](https://pagopa.atlassian.net/browse/PIN-10346)

## 📝 Description / Context

This PR include the fix for some bugs inside drawers:

- bug on the subtitle of the `CustomizeThresholdDrawer`
- bug on the limit alert of the `CustomizeThresholdDrawer`
- bug that show threhsold values in consumer pages components and drawer
- bug that made threshold info in `ProviderThresholdsInfoAlert` to overflow the alert dimension

## 🛠 List of changes

- fix the subtitle to be inline
- fix the alert removing `sx={{ whiteSpace: 'nowrap' }}` that cause the text to overflow the space
- hiding threshold in drawers and sections in consumers pages
- refactor the behaviour of the stack inside the `ProviderThresholdsInfoAlert`

[PIN-10346]: https://pagopa.atlassian.net/browse/PIN-10346?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/shared/CustomizeThresholdDrawer.tsx
+++ b/src/components/shared/CustomizeThresholdDrawer.tsx
@@ -118,9 +118,7 @@ export const CustomizeThresholdDrawer: React.FC<CustomizeThresholdDrawerProps> =
             />
             <GreyAlert>
               <Stack spacing={0.5}>
-                <Typography variant="overline" sx={{ whiteSpace: 'nowrap' }}>
-                  {t('limitAlert.title')}
-                </Typography>
+                <Typography variant="overline">{t('limitAlert.title')}</Typography>
                 <Stack
                   direction={'row'}
                   alignItems={'center'}
--- a/src/components/shared/ProviderThresholdsInfoAlert.tsx
+++ b/src/components/shared/ProviderThresholdsInfoAlert.tsx
@@ -22,7 +22,12 @@ export function ProviderThresholdsInfoAlert({
   return (
     <GreyAlert>
       <AlertTitle sx={{ textTransform: 'uppercase', fontWeight: 700 }}>{t('label')}</AlertTitle>
-      <Stack direction="row" spacing={6} sx={{ mt: 0.5, mb: 1, whiteSpace: 'nowrap' }}>
+      <Stack
+        direction={{ xs: 'column', lg: 'row' }}
+        spacing={{ xs: 1, lg: 6 }}
+        sx={{ mt: 0.5, mb: 1, flexWrap: 'wrap' }}
+        alignItems={{ xs: 'flex-start', lg: 'center' }}
+      >
         <Stack direction="row" spacing={2} alignItems="center">
           <Typography>{t('dailyCallsPerConsumer.label')}</Typography>
           <Typography fontWeight={600}>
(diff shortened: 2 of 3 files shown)


src/pages/ConsumerAgreementCreatePage/components/ConsumerAgreementCreateCertifiedAttributesDrawer.tsx
```diff
@@ -76,6 +76,7 @@ const ConsumerAgreementCreateCertifiedAttributesDrawer: React.FC = () => {
                   checked={isAttributeOwned('certified', attribute.id, ownedCertifiedAttributes, {
                     discreteConfig: attribute.discreteConfig,
                   })}
+                  hideThreshold
```
DenisLantero, line 79: I think this only fixes part of the consumer surface. The catalog consumer page still renders descriptor attributes through `ReadOnlyDescriptorAttributes`, and that component has no way to pass `hideThreshold` down to `AttributeContainer`, so custom per-attribute thresholds can still be shown there when `dailyCallsPerConsumer` is present.

For reference: https://github.com/pagopa/pdnd-interop-frontend/blob/9b0c093187cb83a997d62f2b87f719dd7a66c405/src/pages/ConsumerEServiceDetailsPage/components/ConsumerEServiceDetailsTab/ConsumerEServiceDescriptorAttributes.tsx#L66-L69

Should we make the shared read-only attributes path support hiding thresholds and use it from the consumer catalog view too?

alten-dturus: The `ReadOnlyDescrptorAttributes` fix is in the PR https://github.com/pagopa/pdnd-interop-frontend/pull/2013

DenisLantero (approved):

Alepazz (commented):
The fixes look good, but the new behavior is not covered by tests. The threshold hiding on the consumer certified attribute paths (the create and details drawers and the certified attributes section) has no assertion that an attribute with `dailyCallsPerConsumer` no longer renders the threshold label, and the reworked subtitle in `ProviderEServiceDescriptorAttributesSection` is not exercised because the existing test mocks `CustomizeThresholdDrawer` away. Could we add tests covering these so the fixed behavior is guarded against regressions?

alten-dturus:
> The fixes look good, but the new behavior is not covered by tests. The threshold hiding on the consumer certified attribute paths (the create and details drawers and the certified attributes section) has no assertion that an attribute with `dailyCallsPerConsumer` no longer renders the threshold label, and the reworked subtitle in `ProviderEServiceDescriptorAttributesSection` is not exercised because the existing test mocks `CustomizeThresholdDrawer` away. Could we add tests covering these so the fixed behavior is guarded against regressions?

The test about `hideThreshold` are implemented in another PR inside `AttributeContainer.test`.
The rework about subtitle is only to guarantee that it shows inline

Alepazz (approved):

PR #2011, opened 2026-06-16
fix: fix attribute autocomplete options (PIN-10352)

alten-dturus:
## 🔗 Issue

[PIN-10352](https://pagopa.atlassian.net/browse/PIN-10352)

## 📝 Description / Context

This PR incluse the fix that allow a `CERTIFIED_DISCRETE` attribute to be added more than one time in a descriptor

## 🛠 List of changes

- Change the options on AttributeAutocomplete

[PIN-10352]: https://pagopa.atlassian.net/browse/PIN-10352?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/shared/AddAttributesToForm/AttributeGroup.tsx
+++ b/src/components/shared/AddAttributesToForm/AttributeGroup.tsx
@@ -53,6 +53,7 @@ export const AttributeGroup: React.FC<AttributeGroupProps> = ({
     close: closeConfigureDiscreteAttributeDrawer,
     attribute,
     groupIndex: attributeGroupIndex,
+    attributeIndex,
   } = useConfigureCertifiedDiscreteAttributeDrawer()
 
   const handleDeleteAttributesGroup = () => {
@@ -91,9 +92,11 @@ export const AttributeGroup: React.FC<AttributeGroupProps> = ({
       threshold: threshold,
     }
 
-    if (groups[attributeGroupIndex].some((att) => att.id === attribute.id)) {
-      groups[attributeGroupIndex] = group.map((att) =>
-        att.id === attribute.id ? { ...att, discreteConfig: discreteConfig } : att
+    if (attributeIndex && groups[attributeGroupIndex][attributeIndex].id === attribute.id) {
+      groups[attributeGroupIndex] = group.map((att, index) =>
+        att.id === attribute.id && index === attributeIndex
+          ? { ...att, discreteConfig: discreteConfig }
+          : att
       )
     } else {
       groups[attributeGroupIndex].push({ ...attribute, discreteConfig: discreteConfig })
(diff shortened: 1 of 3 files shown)


src/components/shared/AttributeAutocomplete.tsx
```diff
@@ -86,7 +90,11 @@ export const AttributeAutocomplete: React.FC<AttributeAutocompleteProps> = ({
   const options: { label: string; value: CompactAttribute }[] = React.useMemo(() => {
     const attributes = data?.results ?? []
     return attributes
-      .filter((att) => !alreadySelectedAttributeIds.includes(att.id))
+      .filter(
+        (att) =>
+          (att.kind === 'CERTIFIED' && !alreadySelectedAttributeIds.includes(att.id)) ||
+          att.kind === 'CERTIFIED_DISCRETE'
+      )
```
Alepazz, line 97: When `attributeKey` is `verified` or `declared`, `kindsFilter` already scopes the query to those kinds, so `data.results` holds VERIFIED/DECLARED attributes. This filter only admits CERTIFIED and CERTIFIED_DISCRETE, so it drops every verified and declared option, leaving the autocomplete empty in those tabs of the e-service attributes step. Keep CERTIFIED_DISCRETE always while excluding already-selected for every other kind:

```suggestion
      .filter(
        (att) =>
          att.kind === 'CERTIFIED_DISCRETE' || !alreadySelectedAttributeIds.includes(att.id)
      )
```

src/components/shared/AddAttributesToForm/AttributeGroup.tsx
```diff
@@ -91,9 +92,11 @@ export const AttributeGroup: React.FC<AttributeGroupProps> = ({
       threshold: threshold,
     }
 
-    if (groups[attributeGroupIndex].some((att) => att.id === attribute.id)) {
-      groups[attributeGroupIndex] = group.map((att) =>
-        att.id === attribute.id ? { ...att, discreteConfig: discreteConfig } : att
+    if (attributeIndex && groups[attributeGroupIndex][attributeIndex].id === attribute.id) {
```
Alepazz, line 95: `attributeIndex` is a number, and the first attribute in a group has index `0`, which is falsy. So `attributeIndex && ...` is false for index 0 and the code falls into the `else` branch, pushing a duplicate instead of updating the first attribute's config. The duplicated entry is then mapped into the submit payload. Use an explicit undefined check:

```suggestion
    if (
      attributeIndex !== undefined &&
      groups[attributeGroupIndex][attributeIndex].id === attribute.id
    ) {
```

Alepazz, line 95: Minor, non-blocking: `groups[attributeGroupIndex][attributeIndex]` is accessed by index without checking it exists. If the group changes between opening the drawer and submitting (for example an attribute is removed), `attributeIndex` could be stale and the `.id` access would throw. Worth a defensive guard.

Alepazz (changes requested):
A couple of blocking issues before merging, see inline comments.

alten-dturus:
The task is rejected because the CERTIFIED_DISCRETE attributes should have the same behavior of the CERTIFIED attributes

PR #2012, opened 2026-06-16
fix: fix state problem in UpdateAttributesDrawer (PIN-10353)

alten-dturus:
## 🔗 Issue

[PIN-10353](https://pagopa.atlassian.net/browse/PIN-10353)

## 📝 Description / Context

This PR fix a type inconsistency in UpdateAttributesDrawer and most important fix the problem that the state of the drawer miss a data update and show an old instance of data

## 🛠 List of changes

- Use map function
- Add useEffect at the open to update the data

[PIN-10353]: https://pagopa.atlassian.net/browse/PIN-10353?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/shared/UpdateAttributesDrawer.tsx
+++ b/src/components/shared/UpdateAttributesDrawer.tsx
@@ -46,7 +46,10 @@ export const UpdateAttributesDrawer: React.FC<UpdateAttributesDrawerProps> = ({
   const { eServiceTemplateId, eServiceTemplateVersionId } =
     useParams<'PROVIDE_ESERVICE_TEMPLATE_DETAILS'>()
 
-  const formModelAttributes = mapDescriptorAttributesToFormDescriptorAttributes(attributes)
+  const formModelAttributes = React.useMemo(
+    () => mapDescriptorAttributesToFormDescriptorAttributes(attributes),
+    [attributes]
+  )
 
   const [selectedAttributes, setSelectedAttributes] = React.useState<FormDescriptorAttributes>(() =>
     cloneDeep(formModelAttributes)
@@ -127,7 +130,7 @@ export const UpdateAttributesDrawer: React.FC<UpdateAttributesDrawerProps> = ({
     if (isOpen) {
       setSelectedAttributes(cloneDeep(formModelAttributes))
     }
-  }, [isOpen])
+  }, [formModelAttributes, isOpen])
 
   return (
     <Drawer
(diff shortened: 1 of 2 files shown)


src/components/shared/__tests__/UpdateAttributesDrawer.test.tsx
```diff
@@ -3,6 +3,7 @@ import { render, screen, fireEvent } from '@testing-library/react'
 import { vi, describe, it, expect, beforeEach } from 'vitest'
 import { UpdateAttributesDrawer } from '../UpdateAttributesDrawer'
 import type { FormDescriptorAttributes } from '@/types/attribute.types'
+import { DescriptorAttributes } from '@/api/api.generatedTypes'
```
Alepazz: These two imports are unused after removing the `@/utils/attribute.utils` mock: `FormDescriptorAttributes` is no longer referenced and `DescriptorAttributes` is never used. They raise `no-unused-vars` warnings, so they can be dropped.

```suggestion
```

Alepazz (approved):
LGTM. Left a small inline comment below.

src/components/shared/UpdateAttributesDrawer.tsx
```diff
@@ -118,11 +123,17 @@ export const UpdateAttributesDrawer: React.FC<UpdateAttributesDrawerProps> = ({
     }
   }
 
+  React.useEffect(() => {
+    if (isOpen) {
+      setSelectedAttributes(cloneDeep(formModelAttributes))
+    }
+  }, [isOpen])
```
DenisLantero: Could we clean up this effect dependency before merging? ESLint flags `formModelAttributes` as missing here, but since that value is recomputed on every render, adding it directly would be risky too. 

Maybe the smallest fix is to memoize the mapped attributes from `attributes`, then depend on the memoized value together with `isOpen`.

borgesis95 (approved):

Alepazz:
This activity resolves [PIN-9678](https://pagopa.atlassian.net/browse/PIN-9678) too

[PIN-9678]: https://pagopa.atlassian.net/browse/PIN-9678?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

DenisLantero (approved):
LGTM, just check the inline comment

DenisLantero (approved):

borgesis95 (approved):

Alepazz (approved):

PR #2013, opened 2026-06-16
fix: fix AttributeGroup color and subtitle and AttributeContainer threshold presence (PIN-10356)

alten-dturus:
## 🔗 Issue

[PIN-10356](https://pagopa.atlassian.net/browse/PIN-10356)

## 📝 Description / Context

This PR include the fix that remove threshold in AttributeContainer in consumer page. Also include the fix about the subtitle of the AttributeGroup

## 🛠 List of changes

- set hideThreshold prop
- remove the duplicated subtitle

[PIN-10356]: https://pagopa.atlassian.net/browse/PIN-10356?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/shared/ReadOnlyDescriptorAttributes.tsx
+++ b/src/components/shared/ReadOnlyDescriptorAttributes.tsx
@@ -17,11 +17,7 @@ import type { ActionItemButton } from '@/types/common.types'
 import { attributesHelpLink } from '@/config/constants'
 import { Typography } from '@mui/material'
 import { useCustomizeThresholdDrawer } from './CustomizeThresholdDrawer'
-import {
-  hasAllDescriptorAttributes,
-  isAttributeGroupFullfilled,
-  isAttributeOwned,
-} from '@/utils/attribute.utils'
+import { isAttributeGroupFullfilled, isAttributeOwned } from '@/utils/attribute.utils'
 
 export type AttributeOwnershipData = {
   certified: TenantAttributes['certified']
@@ -57,35 +53,24 @@ export const ReadOnlyDescriptorAttributes: React.FC<ReadOnlyDescriptorAttributes
     )
   }
 
-  const hasBlockingAttribute =
-    !!ownershipData &&
-    !hasAllDescriptorAttributes(
-      'certified',
-      ownershipData.certified,
-      descriptorAttributes.certified
-    )
-
   return (
     <>
       <AttributeGroupsListSection
         descriptorAttributes={descriptorAttributes}
         attributeKey="certified"
         ownershipData={ownershipData}
-        hasBlockingAttribute={hasBlockingAttribute}
       />
       <Divider sx={{ my: 3 }} />
       <AttributeGroupsListSection
         descriptorAttributes={descriptorAttributes}
         attributeKey="verified"
         ownershipData={ownershipData}
-        hasBlockingAttribute={hasBlockingAttribute}
       />
       <Divider sx={{ my: 3 }} />
       <AttributeGroupsListSection
         descriptorAttributes={descriptorAttributes}
         attributeKey="declared"
         ownershipData={ownershipData}
-        hasBlockingAttribute={hasBlockingAttribute}
       />
     </>
   )
@@ -97,7 +82,6 @@ type AttributeGroupsListSectionProps = {
   topSideActions?: Array<ActionItemButton>
   withThreshold?: boolean
   ownershipData?: AttributeOwnershipData
-  hasBlockingAttribute?: boolean
 }
 
 export const AttributeGroupsListSection: React.FC<AttributeGroupsListSectionProps> = ({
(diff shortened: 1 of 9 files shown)


borgesis95 (changes requested):
LGTM, but tests are failing

borgesis95 (approved):

src/components/shared/ReadOnlyDescriptorAttributes.tsx
```diff
@@ -245,7 +245,8 @@ const AttributeGroup: React.FC<AttributeGroupProps> = ({
 
   const shouldHideFulfillmentStatus =
     hasBlockingAttribute && rawGroupColorAndText?.color !== 'error'
-  const groupColorAndText = shouldHideFulfillmentStatus ? undefined : rawGroupColorAndText
+  const groupColor = shouldHideFulfillmentStatus ? undefined : rawGroupColorAndText?.color
+  const groupText = rawGroupColorAndText?.textKey
```
DenisLantero, line 255: I think this keeps leaking fulfillment status text for verified/declared groups when a certified attribute is blocking the whole requirement flow. The card color and checkmark are still hidden through `shouldHideFulfillmentStatus`, but the subtitle now says `group.manage.success.consumer`, which is inconsistent with the disabled/neutral state and reverses the previous test expectation. Should we hide `groupText` it when `shouldHideFulfillmentStatus` is true?

alten-dturus: Now it is right like this because this is what design has choosen as solution when I talk with Barbara about the subtitle we have to show

alten-dturus: Color and icon are hidden if there is a blocking attribute but the text is still visible

DenisLantero: Ok perfect then, keep it as is!

DenisLantero (approved):

src/components/shared/ReadOnlyDescriptorAttributes.tsx
```diff
@@ -287,6 +287,7 @@ const AttributeGroup: React.FC<AttributeGroupProps> = ({
                     : undefined
                 }
                 onCustomizeThreshold={withThreshold ? () => open(attribute, index) : undefined}
+                hideThreshold={!withThreshold}
```
Alepazz: `AttributeGroup` is shared: it also renders through `AttributeGroupsListSection` on the provider side, where the e-service summary and template views pass no `withThreshold`, and the descriptor details view passes it only to the certified section. `hideThreshold` was never set before, so `dailyCallsPerConsumer` used to render in those provider views, and `hideThreshold={!withThreshold}` now hides it there too. Is suppressing the threshold intended in every view without `withThreshold`, or should it be scoped to the consumer read-only page? Verified and declared groups are unaffected (no threshold), so this only concerns certified attributes in the provider summary and template.

src/components/shared/__tests__/ReadOnlyDescriptorAttributes.test.tsx
```diff
@@ -160,11 +160,8 @@ describe('ReadOnlyDescriptorAttributes', () => {
       // The certified group should still show the error text
       expect(screen.getByText('group.manage.error.consumer')).toBeInTheDocument()
 
-      // Verified and declared groups should hide fulfillment status (no subtitle shown)
-      expect(screen.queryByText('consumer')).not.toBeInTheDocument()
-
-      // Success text should NOT appear (verified is fulfilled but hidden due to blocking attribute)
-      expect(screen.queryByText('group.manage.success.consumer')).not.toBeInTheDocument()
+      // Success text should appear (verified and declared are fulfilled and only color is hidden due to blocking attribute)
+      expect(screen.queryAllByText('group.manage.success.consumer').length).toBe(2)
```
Alepazz, line 197: A few gaps in how the new behavior is covered here:
- This asserts the success text count but not that the non-blocking groups now render with the neutral gray color, so a regression that kept the previous success or warning color would still pass.
- The threshold hiding, which is the main fix, is never exercised: these attributes have no `dailyCallsPerConsumer`, and `AttributeContainer` has no test for `hideThreshold`, so neither "threshold hidden in read-only" nor "threshold still shown when `withThreshold` is set" is protected.
- Unrelated to this change but worth fixing while here: the `should hide checkmarks` test queries `CheckIcon`, while `AttributeContainer` renders `CheckCircleIcon`, so it matches nothing and would pass even if a checkmark were rendered.

Alepazz (commented):
A few questions and some test-coverage notes inline.

src/components/shared/ReadOnlyDescriptorAttributes.tsx
alten-dturus: From the design, the only times where the threshold are visible is when they can be modified. Maybe there is a design mistake or miss

alten-dturus: I think design has not explored every case where this component is used and so in the design many cases are not present

Alepazz (approved):

PR #2015, opened 2026-06-16
fix: risk analysis put service types (PIN-10364)

l-giacomini:
## Issue
- [PIN-10364](https://pagopa.atlassian.net/browse/PIN-10364)
- Fixed PUT /riskAnalysis/form type and data passed
- Updated tests

[PIN-10364]: https://pagopa.atlassian.net/browse/PIN-10364?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/api/purpose/purpose.services.ts
+++ b/src/api/purpose/purpose.services.ts
@@ -339,11 +339,10 @@ async function updateRiskAnalysis({
   purposeId,
   ...payload
 }: { purposeId: string } & RiskAnalysisFormSeed) {
-  const response = await axiosInstance.put<PurposeVersionResource>(
+  return axiosInstance.put(
     `${BACKEND_FOR_FRONTEND_URL}/purposes/${purposeId}/riskAnalysis/form`,
     payload
   )
-  return response.data
 }
 
 async function getRiskAnalysisAssignments(params: GetRiskAnalysisAssignmentsParams) {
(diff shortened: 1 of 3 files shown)


src/api/purpose/purpose.services.ts
```diff
@@ -337,7 +338,7 @@ async function rejectRiskAnalysis({
 async function updateRiskAnalysis({
   purposeId,
   ...payload
-}: { purposeId: string } & PurposeUpdateContent) {
+}: { purposeId: string } & RiskAnalysisFormSeed) {
   const response = await axiosInstance.put<PurposeVersionResource>(
```
Alepazz: The request body type is now correct. While here, the response is still typed as `PurposeVersionResource` and the function returns `response.data`, but the contract for this endpoint declares the response as `void`. No caller reads the return value, so it is harmless today, but aligning the response type to `void` (and dropping the `return response.data`) would complete the type fix for this endpoint.

Alepazz (approved):
LGTM, left a small inline comment below.

src/api/purpose/purpose.services.ts
l-giacomini: Fixed in: https://github.com/pagopa/pdnd-interop-frontend/pull/2015/changes/4149adb77c816d8d758683dfd5938398c5524a64

Alepazz (approved):

PR #2018, opened 2026-06-17
fix: update notification banner S3 path (PIN-10357)

DenisLantero:
## 🔗 Issue

[PIN-10357](https://pagopa.atlassian.net/browse/PIN-10357)

## 📝 Description / Context

Updates the notification banner configuration URL to read the S3 resource from the new `banner-info-window` path.

## 🛠 List of changes

- Replaced the notification banner S3 path from `/notifications-window/data.json` to `/banner-info-window/data.json`.

## 🧪 How to test

- Open the application in an environment where the notification banner configuration is fetched from S3.
- Verify that the request targets `<INTEROP_RESOURCES_BASE_URL>/banner-info-window/data.json`.

[PIN-10357]: https://pagopa.atlassian.net/browse/PIN-10357?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/api/notification/notification.services.ts
+++ b/src/api/notification/notification.services.ts
@@ -79,7 +79,7 @@ async function getNotificationsBannerConfigJson() {
   }
 
   const response = await axios.get<BannerData>(
-    `${INTEROP_RESOURCES_BASE_URL}/notifications-window/data.json`
+    `${INTEROP_RESOURCES_BASE_URL}/banner-info-window/data.json`
   )
   return response.data
 }
--- a/src/api/notification/notification.services.ts
+++ b/src/api/notification/notification.services.ts
@@ -1,6 +1,5 @@
 import axiosInstance from '@/config/axios'
-import axios from 'axios'
-import { APP_MODE, BACKEND_FOR_FRONTEND_URL, INTEROP_RESOURCES_BASE_URL } from '@/config/env'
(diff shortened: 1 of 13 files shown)


src/api/notification/notification.services.ts
```diff
@@ -79,7 +79,7 @@ async function getNotificationsBannerConfigJson() {
   }
 
   const response = await axios.get<BannerData>(
-    `${INTEROP_RESOURCES_BASE_URL}/notifications-window/data.json`
```
borgesis95, line 82: Could we move out this API in other services and queries file? R Is not anymore associated with notification feature

borgesis95 (changes requested):

Alepazz (approved):
LGTM. One minor, non-blocking note: `productUpdates.services.ts` imports `BannerData` from the hooks layer (`@/hooks/bannerHooks/utils`), so the new API module ends up depending on the hooks layer. It carries over from the previous code, but since this is a brand new module it could be a good spot to move the type to a more neutral place (for example the api types).

borgesis95 (changes requested):
Can I ask you to address last @Alepazz comment?

PR #2023, opened 2026-06-18
chore/PIN-9981: changed purpose isFreeOfCharge into a switch

emabruz94:
## 🔗 Issue

[PIN-9981](https://pagopa.atlassian.net/browse/PIN-9981)

## 📝 Description / Context

This PR updates the UX of the new Purpose creation flow (**Submitted Purposes → Create New**) by replacing the radio button selection with a switch for the option related to making the data received from the e-service available free of charge.

Specifically, the "free of charge" option is no longer managed through radio buttons and is now represented by a switch:

* **Switch enabled:** the Purpose is free of charge, and an additional section (reason for free-of-charge availability) is displayed.
* **Switch disabled:** behavior is equivalent to the paid option.

This change introduces a more intuitive interaction pattern while preserving the existing business logic and submission behavior.

## 🛠 List of changes

* Replaced the free-of-charge radio button selection with a switch in the new Purpose creation flow.
* Added conditional rendering of the related details section when the switch is enabled.
* Preserved payload structure and submission behavior according to the selected switch state.
* Updated labels and form interactions to align with the new UX pattern.

## 🧪 How to test

1. Open the flow: **Submitted Purposes → Create New**.
2. Navigate to the step where users choose how the data received from the e-service will be made available.
3. Verify that a switch is displayed instead of the previous radio buttons.
4. Enable the switch.
5. Verify that the additional section for the free-of-charge reason appears below the field.
6. Disable the switch.
7. Verify that the additional section is hidden.
8. Complete and save the Purpose in both scenarios (switch enabled and disabled) and verify that the final behavior is consistent with the selected option.
9. Perform a quick regression check on the Purpose edit and summary views to confirm that the selected value is correctly persisted and displayed.

[PIN-9981]: https://pagopa.atlassian.net/browse/PIN-9981?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/pages/ConsumerPurposeEditPage/components/PurposeEditStepGeneral/__tests__/PurposeEditStepGeneralForm.test.tsx
+++ b/src/pages/ConsumerPurposeEditPage/components/PurposeEditStepGeneral/__tests__/PurposeEditStepGeneralForm.test.tsx
@@ -29,13 +29,13 @@ vi.mock('@/components/shared/PurposeLoadEstimationSection', () => ({
 const defaultValues: PurposeEditStepGeneralFormValues = {
   title: 'Test purpose',
   description: 'A test purpose description',
-  isFreeOfCharge: 'NO',
+  isFreeOfCharge: false,
   dailyCalls: 100,
 }
 
 function renderComponent(overrides?: {
   purposeMode?: 'DELIVER' | 'RECEIVE'
-  isFreeOfCharge?: 'YES' | 'NO'
+  isFreeOfCharge?: boolean
 }) {
   const purpose = createMockPurpose({
     eservice: {
@@ -67,7 +67,7 @@ function renderComponent(overrides?: {
 
 describe('PurposeEditStepGeneralForm', () => {
   it('should not show freeOfChargeReason field when isFreeOfCharge is NO', () => {
-    renderComponent({ isFreeOfCharge: 'NO' })
+    renderComponent({ isFreeOfCharge: false })
 
     expect(
       screen.queryByRole('textbox', { name: 'edit.stepGeneral.freeOfChargeReasonField.label' })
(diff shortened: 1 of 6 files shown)


Alepazz (changes requested):
Please take a look at the open Copilot threads, in particular the one about the default freeOfChargeReason value being sent as backend data rather than used as UI copy.

emabruz94:
The Copilot issue about the default value is a known and wanted one, this PR didn't change its behavior; the other issues have already been resolved

Alepazz (approved):
Appreciate the clarification, the design you shared confirms the new copy is intended, so the default freeOfChargeReason point is settled for me. Approving.

As a suggestion, not blocking: `PurposeEditStepGeneralFormValues` still omits `isFreeOfCharge` and re-adds it as `boolean`, and re-declares `dailyCalls`, which is now redundant since `PurposeUpdateContent` already types both. Worth tidying up whenever convenient.

src/pages/ConsumerPurposeEditPage/components/PurposeEditStepGeneral/__tests__/PurposeEditStepGeneralForm.test.tsx
```diff
@@ -111,4 +116,159 @@ describe('PurposeEditStepGeneralForm', () => {
 
     expect(screen.getByRole('button', { name: 'edit.endWithSaveBtn' })).toBeInTheDocument()
   })
+
+  it('submits with DELIVER mode and calls updateDraft with riskAnalysisForm', async () => {
```
DenisLantero, line 120: Could we add the matching submit case with the switch OFF? The PR behavior explicitly covers saving both enabled and disabled states, but the submit tests only assert the `true` path. A focused test can start from `isFreeOfCharge: false`, submit, and expect `isFreeOfCharge: false` plus no `freeOfChargeReason` in the payload.

DenisLantero (approved):
LGTM, check the non-blocking inline comment

PR #2025, opened 2026-06-18
fix/PIN-10160: navigate with purposeId to ConsumerDebugVoucherPage

emabruz94:
## 🔗 Issue

[PIN-10160](https://pagopa.atlassian.net/browse/PIN-10160)

## 📝 Description / Context

This PR fixes a navigation issue affecting the Voucher Debug page when the `purposeId` parameter is missing.

In the `bugfix/PIN-10160_missing-purposeId` branch, the navigation to `ConsumerDebugVoucherPage` has been updated to ensure that the `purposeId` is correctly passed through the voucher simulation flow. This prevents the debug page from opening in an incomplete state or with data that is inconsistent with the selected Purpose.

## 🛠 List of changes

* Fixed the propagation of the `purposeId` parameter when navigating to the Voucher Debug page.
* Aligned the voucher simulation flow to preserve the selected Purpose context during redirects to the debug page.
* Reduced the risk of errors caused by missing query parameters during the debug/simulation process.

## 🧪 How to test

1. Start the application locally.
2. Navigate to the consumer voucher simulation flow and select a valid Purpose.
3. From the screen that exposes the Voucher Debug action, click the link/button that navigates to the debug page.
4. Verify that the Voucher Debug page opens while preserving the context of the selected Purpose (`purposeId` is present in the URL parameters and the expected data is correctly loaded).
5. Repeat the test using the different entry points of the flow (where the debug link/button is available) to verify that no regressions have been introduced.

[PIN-10160]: https://pagopa.atlassian.net/browse/PIN-10160?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/pages/ConsumerClientManagePage/ConsumerClientManage.page.tsx
+++ b/src/pages/ConsumerClientManagePage/ConsumerClientManage.page.tsx
@@ -25,7 +25,7 @@ const ConsumerClientManagePage: React.FC = () => {
   const { t: tCommon } = useTranslation('common', { keyPrefix: 'actions' })
   const { clientId } = useParams<'SUBSCRIBE_CLIENT_EDIT' | 'SUBSCRIBE_INTEROP_M2M_CLIENT_EDIT'>()
   const [searchParams] = useSearchParams()
-  const purposeId = searchParams.get('purposeId') || ''
+  const purposeId = searchParams.get('purposeId')
   const clientKind = useClientKind()
   const { activeTab, updateActiveTab } = useActiveTab('clientOperators')
   const navigate = useNavigate()
--- a/src/pages/ConsumerClientManagePage/ConsumerClientManage.page.tsx
+++ b/src/pages/ConsumerClientManagePage/ConsumerClientManage.page.tsx
@@ -1,6 +1,6 @@
 import { ClientMutations, ClientQueries } from '@/api/client'
 import { PageContainer, SectionContainer } from '@/components/layout/containers'
-import { useParams } from '@/router'
+import { useParams, useNavigate } from '@/router'
 import { useActiveTab } from '@/hooks/useActiveTab'
 import { useMarkNotificationsAsRead } from '@/hooks/useMarkNotificationsAsRead'
 import { TabContext, TabList, TabPanel } from '@mui/lab'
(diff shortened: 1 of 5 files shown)


src/pages/ConsumerClientManagePage/ConsumerClientManage.page.tsx
```diff
@@ -17,13 +17,15 @@ import DeleteIcon from '@mui/icons-material/DeleteOutline'
...
-import { useNavigate } from '@/router'
 import type { ActionItemButton } from '@/types/common.types'
+import { useSearchParams } from 'react-router-dom'
 
 const ConsumerClientManagePage: React.FC = () => {
   const { t } = useTranslation('client', { keyPrefix: 'edit' })
   const { t: tCommon } = useTranslation('common', { keyPrefix: 'actions' })
   const { clientId } = useParams<'SUBSCRIBE_CLIENT_EDIT' | 'SUBSCRIBE_INTEROP_M2M_CLIENT_EDIT'>()
+  const [searchParams] = useSearchParams()
+  const purposeId = searchParams.get('purposeId') || ''
```
Alepazz: `searchParams.get` returns `null` when the param is absent; `|| ''` converts it to an empty string, which can produce `?purposeId=` in the URL. The voucher simulation form reads `purposeId` via `searchParams.get('purposeId')` without a fallback, so `null` and `''` have different effects on the form's initial state (the purpose select initializes differently for a missing param vs. an explicit empty one).

```suggestion
  const purposeId = searchParams.get('purposeId')
```

src/pages/ConsumerClientManagePage/ConsumerClientManage.page.tsx
```diff
@@ -50,7 +52,7 @@ const ConsumerClientManagePage: React.FC = () => {
     action: () =>
       navigate(
         clientKind === 'API' ? 'SIMULATE_GET_VOUCHER_API' : 'SIMULATE_GET_VOUCHER_CONSUMER',
-        { urlParams: { clientId } }
+        { urlParams: { clientId, purposeId } }
```
Alepazz: With `purposeId` now typed as `string | null`, you'll also want to spread it conditionally so it's only added to `urlParams` when actually present, otherwise `null` can still end up as an unexpected param value.

```suggestion
        { urlParams: { clientId, ...(purposeId ? { purposeId } : {}) } }
```

**Missing test coverage:** the new behavior (forwarding `purposeId` from the client page to the voucher simulation URL) has no automated coverage. A test verifying that the navigate action includes `purposeId` when present, and omits it when absent, would lock in this contract.

Alepazz (changes requested):
One concern about the empty-string fallback, see inline comment.

Alepazz (approved):
Fix looks correct, approving.
One request before merge, if you can: please add the test mentioned earlier, verifying that the voucher simulation navigation includes purposeId when it is present and omits it when it is absent. It is small, but it locks in this contract and matches how the surrounding voucher flow is already covered.

DenisLantero (approved):
LGTM, could you add the missing test coverage like @Alepazz said?

PR #2027, opened 2026-06-18
fix: show "Crea finalità" title during purpose draft creation (PIN-7610)

Alepazz:
## 🔗 Issue
[PIN-7610](https://pagopa.atlassian.net/browse/PIN-7610)

## 📝 Description / Context
When creating a purpose, after filling in the consumer and the e-service and clicking "Crea bozza", the user lands on the purpose edit step: the draft already exists technically, but the key information has not been entered yet. In this phase the page title showed "Modifica finalità", while it should read "Crea finalità". The edit page is reached both from the creation flow (first edit) and from re-editing an existing draft, and nothing distinguished the two cases: the purpose is in `DRAFT` state in both, and its data is filled in progressively during the flow, so there is no reliable data-only signal. The discriminator is therefore passed as navigation state (`isFirstEdit`), the same idiom already used elsewhere in the app (e.g. `stepIndexDestination`). The "Crea finalità" copy already exists in the locale files (`create.emptyTitle`), so no new copy was introduced.

## 🛠 List of changes
- `PurposeCreateForm`: pass `state: { isFirstEdit: true }` when navigating to the edit page after a successful draft creation (both the DELIVER and the RECEIVE flows).
- `ConsumerPurposeEditPage`: read `useLocation().state.isFirstEdit` and render `create.emptyTitle` ("Crea finalità") during the creation phase, `edit.emptyTitle` ("Modifica finalità") otherwise.
- Tests: page-level regression test on the conditional title, plus a create-form test asserting the navigation state is sent on both creation flows.

## 🧪 How to test
- Go to Fruizione → Finalità → "Crea finalità", select a consumer and an e-service, then click "Crea bozza".
- On the resulting edit step the page title reads "Crea finalità" (see attached screenshot).
- Open an existing draft from the purpose summary ("modifica"): the title reads "Modifica finalità".

## ⚠️ Known limitations / out of scope
- On a hard refresh of the edit step the navigation state is lost, so the title falls back to "Modifica finalità". There is no reliable data-only way to tell "just created" from "re-editing a draft" (both are `DRAFT`).
- The breadcrumb still reads "Modifica finalità": it is a route-derived label rendered by a global shared component, not driven by navigation state, so it is left out of scope (the ticket targets the page title).
- The "purpose from template" edit page (`ConsumerPurposeFromTemplateEditPage`) has the same hardcoded title and is intentionally not addressed here, leaving it for the "template finalità" feature as noted in the ticket.

---

PIN-7610-crea-finalita-title

[PIN-7610]: https://pagopa.atlassian.net/browse/PIN-7610

### DIFF
--- a/src/pages/ConsumerPurposeCreatePage/components/PurposeCreateForm.tsx
+++ b/src/pages/ConsumerPurposeCreatePage/components/PurposeCreateForm.tsx
@@ -185,6 +185,7 @@ export const PurposeCreateForm: React.FC<PurposeCreateFormProps> = ({ purposeTem
           const purposeId = data.id
           navigate('SUBSCRIBE_PURPOSE_EDIT', {
             params: { purposeId },
+            state: { isFirstEdit: true },
           })
         },
       })
@@ -207,6 +208,7 @@ export const PurposeCreateForm: React.FC<PurposeCreateFormProps> = ({ purposeTem
           const purposeId = data.id
           navigate('SUBSCRIBE_PURPOSE_EDIT', {
             params: { purposeId },
+            state: { isFirstEdit: true },
           })
         },
       })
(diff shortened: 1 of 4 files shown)


antofo (approved):
The PR is ready to merge. Like you wrote in the description, there are issues linked to this one that are not addressed  because not mentioned in the ticket and so out-of-scope. I think in the future other bugs will be open, I'm unsure whether it's appropriate to manage them preemptively or wait.

borgesis95 (changes requested):
It works but (as you described on PR description) it's so fragile. we need to work on all stuff that within page indicate create/update (my fault ticket doesn't tell about it). I let "request changes" in order to think deeper about this ticket.

PR #2028, opened 2026-06-18
bugfix: debug voucher client id validation (PIN-10328)

DenisLantero:
## 🔗 Issue

[PIN-10328](https://pagopa.atlassian.net/browse/PIN-10328)

## 📝 Description / Context

This PR prevents invalid `client_id` / `producer_id` values from being submitted from the Client Assertion Debugger when they are not valid UUIDs.

## 🛠 List of changes

- Added a shared UUID validation helper based on `zod`.
- Added frontend validation to the debug voucher `client_id or producer_id` input.
- Added localized validation messages.
- Added tests for the shared UUID helper and the form validation behavior.

## 🧪 How to test

- Open the Client Assertion Debugger page.
- Fill in the required Client Assertion field.
- Enter a non-UUID value in the `client_id or producer_id` field and submit: the form should show an inline validation error and should not send the request.
- Enter a valid UUID and submit: the request should be sent normally.

[PIN-10328]: https://pagopa.atlassian.net/browse/PIN-10328?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sandrotaje (commented):
It looks good to me but I think in the ticket was asked also to enforce dpop and client assertion to be a valid JWT. Is it planned in a different PR?

DenisLantero:
> It looks good to me but I think in the ticket was asked also to enforce dpop and client assertion to be a valid JWT. Is it planned in a different PR?

Good catch, I read this and didn't notice it later said to implement the pattern on the JWT inputs as well:

> In ambiente di collaudo, per il Client Assertion Debugger se non si inserisce un formato valido di UUID per ClientId OR ProducerID  il debugger si rompe, vedi snapshot:

I added the fix [here](https://github.com/pagopa/pdnd-interop-frontend/pull/2028/commits/645d16fcc1cb2a5a2234cc8ef08664db1732f490)

sandrotaje (approved):

LucaPierfederici (approved):
LGTM, one small change:
The PR description only covers the UUID validation on `client_id / producer_id`, but the code also adds JWT validation on `Client assertion` and `DPoP Proof`. 
Could you update the PR description to include the JWT part too?

src/utils/form.utils.ts
```diff
@@ -66,6 +67,17 @@ export const emailRegex =
 export const urlRegex =
   /^https?:\/\/(?:www\.)?[-a-zA-Z0-9@:%._\+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b(?:[-a-zA-Z0-9()@:%_\+.~#?&\/=]*)$/
 
+const uuidSchema = z.string().uuid()
```
Alepazz: Optional: `z.string().uuid()` is the deprecated chained form in Zod 4, the top-level `z.uuid()` is the recommended one and would line up with the `z.jwt()` used right below.

```suggestion
const uuidSchema = z.uuid()
```

src/pages/ConsumerDebugVoucherPage/components/__test__/DebugVoucherForm.test.tsx
```diff
@@ -13,15 +13,21 @@ import {
 import type { TokenGenerationValidationResult } from '@/api/api.generatedTypes'
 
 const response = createMockDebugVoucherResultPassed()
+const validateTokenGenerationHandler = vi.fn()
+const validJwt = 'eyJhbGciOiJSUzI1NiJ9.eyJzdWIiOiJ0ZXN0In0.c2lnbmF0dXJl'
```
Alepazz: Optional: this `validJwt` constant is duplicated verbatim in the page-level test too. Lifting it into a shared test fixture would keep the two in sync.

Alepazz (approved):
LGTM. A couple of optional nits below.

PR #2029, opened 2026-06-18
fix: hide keychain tab from API operators (PIN-9909)

DenisLantero:
## 🔗 Issue

[PIN-9909](https://pagopa.atlassian.net/browse/PIN-9909)

## 📝 Description / Context

API operators could see and open the producer Keychain tab from the provider EService details page, even though the backend does not allow them to access that area.

## 🛠 List of changes

- Hide the Keychain tab and panel for non-admin users.
- Keep non-admin users on the EService details tab when the URL requests `?tab=keychains`.
- Remove the keychain alert navigation action when the user cannot access the Keychain tab.
- Add focused tests for admin visibility and API operator restrictions.

## 🧪 How to test

- Open a provider EService details page as an admin: the Keychain tab should be visible and accessible.
- Open the same page as an API operator: the Keychain tab should not be visible.
- As an API operator, open the page with `?tab=keychains`: the EService details tab should remain selected and the Keychain content should not render.

[PIN-9909]: https://pagopa.atlassian.net/browse/PIN-9909?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/pages/ProviderEServiceDetailsPage/ProviderEServiceDetails.page.tsx
+++ b/src/pages/ProviderEServiceDetailsPage/ProviderEServiceDetails.page.tsx
@@ -20,12 +20,16 @@ import {
   isDescriptorPendingArchiving,
 } from '@/utils/eservice.utils'
 import { ProviderEServiceDetailsAlerts } from './components/ProviderEServiceDetailsTab/ProviderEServiceDetailsAlerts'
+import { AuthHooks } from '@/api/auth'
 
 const ProviderEServiceDetailsPage: React.FC = () => {
   const { t } = useTranslation('eservice', { keyPrefix: 'read' })
   const { eserviceId, descriptorId } = useParams<'PROVIDE_ESERVICE_MANAGE'>()
 
   const { activeTab, updateActiveTab } = useActiveTab('eserviceDetails')
+  const { isAdmin } = AuthHooks.useJwt()
+  const canViewKeychains = isAdmin
+  const selectedTab = canViewKeychains ? activeTab : 'eserviceDetails'
   const { openDialog } = useDialog()
   const {
     isOpen: isVersionSelectorDrawerOpen,
@@ -47,6 +51,7 @@ const ProviderEServiceDetailsPage: React.FC = () => {
   )
 
   const handleViewKeychains = () => {
+    if (!canViewKeychains) return
     updateActiveTab(null, 'keychains')
   }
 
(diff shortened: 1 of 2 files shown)


src/pages/ProviderEServiceDetailsPage/ProviderEServiceDetails.page.tsx
```diff
@@ -20,12 +20,16 @@ import {
...
 import { ProviderEServiceDetailsAlerts } from './components/ProviderEServiceDetailsTab/ProviderEServiceDetailsAlerts'
+import { AuthHooks } from '@/api/auth'
 
 const ProviderEServiceDetailsPage: React.FC = () => {
   const { t } = useTranslation('eservice', { keyPrefix: 'read' })
   const { eserviceId, descriptorId } = useParams<'PROVIDE_ESERVICE_MANAGE'>()
 
   const { activeTab, updateActiveTab } = useActiveTab('eserviceDetails')
+  const { isAdmin } = AuthHooks.useJwt()
+  const canViewKeychains = isAdmin
```
Alepazz: `isAdmin` here is true only for a pure admin (`currentRoles.length === 1 && currentRoles[0] === 'admin'`), so this hides the keychain tab from `support` and `security` users too, not just API operators. That conflicts with the role model elsewhere: the producer keychain routes are readable by `['admin', 'support', 'security']`, and `ProviderKeychainsListPage` reserves only the create action to admins while still rendering the keychains table for the other read roles. Since the ticket is scoped to API operators, consider aligning this with the keychain routes' auth levels, for example `isAdmin || isSupport || isOperatorSecurity`, so support and security keep their read access. The new tests only cover admin vs API operator, so this case isn't currently caught.

src/pages/ProviderEServiceDetailsPage/ProviderEServiceDetails.page.tsx
```diff
@@ -47,6 +51,7 @@ const ProviderEServiceDetailsPage: React.FC = () => {
   )
 
   const handleViewKeychains = () => {
+    if (!canViewKeychains) return
```
Alepazz: Since `onViewKeychains` is only passed when `canViewKeychains` is true, this early return is unreachable. One of the two guards can be dropped.

src/pages/ProviderEServiceDetailsPage/ProviderEServiceDetails.page.tsx
```diff
@@ -20,12 +20,16 @@ import {
...
+import { AuthHooks } from '@/api/auth'
 
 const ProviderEServiceDetailsPage: React.FC = () => {
   const { t } = useTranslation('eservice', { keyPrefix: 'read' })
   const { eserviceId, descriptorId } = useParams<'PROVIDE_ESERVICE_MANAGE'>()
 
   const { activeTab, updateActiveTab } = useActiveTab('eserviceDetails')
+  const { isAdmin } = AuthHooks.useJwt()
+  const canViewKeychains = isAdmin
+  const selectedTab = canViewKeychains ? activeTab : 'eserviceDetails'
```
Alepazz, line 32: For a restricted user landing on `?tab=keychains`, `selectedTab` correctly forces the details panel, but the `tab` query param stays `keychains` in the URL. It's only a cosmetic stale-state mismatch since the right panel renders, but normalizing the param would keep URL and view in sync.

Alepazz (changes requested):
See inline comments.

src/pages/ProviderEServiceDetailsPage/ProviderEServiceDetails.page.tsx
DenisLantero: Fixed [here](https://github.com/pagopa/pdnd-interop-frontend/pull/2029/commits/eefcc29ac60617a06ed45aa1e98095c8155c96ac)

src/pages/ProviderEServiceDetailsPage/ProviderEServiceDetails.page.tsx
DenisLantero: Fixed [here](https://github.com/pagopa/pdnd-interop-frontend/pull/2029/commits/c5dfd64e1741470b32312b2da57aafc39095c5e4)

src/pages/ProviderEServiceDetailsPage/ProviderEServiceDetails.page.tsx
DenisLantero: Fixed [here](https://github.com/pagopa/pdnd-interop-frontend/pull/2029/commits/4dc19297cfab1f29148a9eac1cb4a46e714b646c)

Alepazz (approved):
Great Job!

GorlemZ (approved):

PR #2030, opened 2026-06-19
fix: use local maintenance font stylesheet (PIN-7939)

DenisLantero:
## 🔗 Issue

[PIN-7939](https://pagopa.atlassian.net/browse/PIN-7939)

## 📝 Description / Context

Maintenance pages no longer depend on the SelfCare hosted font stylesheet with a fixed SRI hash, avoiding failures when the remote CSS changes.

## 🛠 List of changes

- Added a local versioned `maintenance-page/fonts.css` stylesheet.
- Updated both maintenance HTML pages to load the local font stylesheet.
- Added a guard test for maintenance page font assets.

## 🧪 How to test

- Open `maintenance-page/maintenance-standard.html` and `maintenance-page/maintenance-token-down.html` and verify the pages still use Titillium Web styling.
- Inspect the page source and verify the font stylesheet is loaded from `./fonts.css`, without the old SelfCare CSS `integrity` link.

[PIN-7939]: https://pagopa.atlassian.net/browse/PIN-7939?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/maintenance-page/fonts.css
+++ b/maintenance-page/fonts.css
@@ -36,36 +36,3 @@
     url('https://selfcare.pagopa.it/assets/font/Titillium_Web/TitilliumWeb-Bold.ttf')
       format('truetype');
 }
-
-@font-face {
-  font-family: 'Titillium Sans Pro';
-  font-display: swap;
-  font-weight: 400;
-  src:
-    url('https://selfcare.pagopa.it/assets/font/Titillium_Sans_Pro/TitilliumSansPro-Regular.woff2')
-      format('woff2'),
-    url('https://selfcare.pagopa.it/assets/font/Titillium_Sans_Pro/TitilliumSansPro-Regular.woff')
-      format('woff');
-}
-
-@font-face {
-  font-family: 'Titillium Sans Pro';
-  font-display: swap;
-  font-weight: 600;
-  src:
-    url('https://selfcare.pagopa.it/assets/font/Titillium_Sans_Pro/TitilliumSansPro-SemiBold.woff2')
-      format('woff2'),
-    url('https://selfcare.pagopa.it/assets/font/Titillium_Sans_Pro/TitilliumSansPro-SemiBold.woff')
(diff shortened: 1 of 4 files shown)


maintenance-page/fonts.css
```diff
@@ -0,0 +1,71 @@
...
+      format('truetype');
+}
+
+@font-face {
+  font-family: 'Titillium Sans Pro';
+  font-display: swap;
+  font-weight: 400;
+  src:
+    url('https://selfcare.pagopa.it/assets/font/Titillium_Sans_Pro/TitilliumSansPro-Regular.woff2')
+      format('woff2'),
```
GorlemZ: These three `Titillium Sans Pro` @font-face blocks don't seem to be used anywhere, they're ~half of the new file and add 6 more remote URLs.

GorlemZ (commented):
Looks good for what it targets -> the SRI-hash failure on the remote CSS is gone and
the change is low-risk. A simple follow up: the font binaries are still loaded from selfcare.pagopa.it, so this is a partial self-hosting. Have you considered this?

DenisLantero:
> Looks good for what it targets -> the SRI-hash failure on the remote CSS is gone and the change is low-risk. A simple follow up: the font binaries are still loaded from selfcare.pagopa.it, so this is a partial self-hosting. Have you considered this?

Good point, I kept it scoped to the CSS file because the breakage came from the `integrity` filed. 
If SelfCare chagnes that file, the hash no longer matches the one that we had in the maintenance pages, so the browser refused to load the stylesheet.

I don't think the scope of the issue was to remove the dependency with selfcare on fonts, I think it was just to avoid this issue.

antofo (approved):

borgesis95 (approved):

PR #2036, opened 2026-06-22
fix: PIN-10396 offer version upgrade instead of new fruition request on obsolete version

Alepazz:
## 🔗 Issue

[PIN-10396](https://pagopa.atlassian.net/browse/PIN-10396)

## 📝 Description / Context

In the consumer e-service catalog detail, "Richiedi fruizione" was enabled even when the requester already had an agreement for that e-service on an obsolete descriptor version, so clicking it returned a `409 Agreement already exists`. Root cause: the catalog descriptor payload of the active version does not expose agreements that live on older descriptor versions (`agreements: []`, `isSubscribed: false`), so the action logic computed "no agreement → can subscribe". The signal that an upgradeable agreement exists lives only in the `GET /consumers/agreements` response (`canBeUpgraded`). Per the "Fruitore di versione obsoleta" design, in this state the correct primary action is "Aggiorna a nuova versione", reusing the existing agreement upgrade flow, instead of a new fruition request.

## 🛠 List of changes

- `ConsumerEServiceDetails.page.tsx`: fetch the requester's agreements for the e-service, derive whether a blocking agreement exists and which one is upgradeable, and (only when needed) fetch the full agreement plus attribute ownership for the upgrade dialog; the subscribe action is suppressed while that query is still loading to avoid a brief wrong state.
- `useGetEServiceConsumerActions.ts`: new `requesterEserviceAgreement` input that suppresses "Richiedi fruizione" when a blocking agreement exists and adds the "Aggiorna a nuova versione" action (reuses the `upgradeAgreementVersion` dialog and the `Update` icon, disabled with a tooltip when certified attributes are missing); the subscribe action is preserved for delegators without an agreement.
- `agreement.utils.ts`: new `getRequesterObsoleteVersionAgreement` helper that flags a blocking agreement and picks the upgradeable one only when its state is `ACTIVE`/`SUSPENDED`, consistent with the agreements list.
- New i18n key `tableEServiceCatalog.upgradeToNewVersion` (it/en).
- Unit tests for the new hook branches and the helper.

## 🧪 How to test

- As a consumer that already holds an active agreement on an older version of an e-service, open that e-service from the catalog on its active (newer) version: the primary action is "Aggiorna a nuova versione" and clicking it opens the upgrade dialog (no more `409`).
- For an e-service where you have no agreement, "Richiedi fruizione" still appears as before.
- For a delegator without an agreement, the subscribe action still shows as a secondary button alongside the upgrade action.

---
PIN-10396-upgrade-action

[PIN-10396]: https://pagopa.atlassian.net/browse/PIN-10396

### DIFF
--- a/src/hooks/useGetEServiceConsumerActions.ts
+++ b/src/hooks/useGetEServiceConsumerActions.ts
@@ -22,19 +22,21 @@ import noop from 'lodash/noop'
 import { useDialog } from '@/stores'
 import { match } from 'ts-pattern'
 
+export type RequesterEserviceAgreement = {
+  blocksSubscribe: boolean
+  upgrade?: {
+    agreement: Agreement
+    hasMissingAttributes: boolean
+    hasAllCertifiedAttributes: boolean
+  }
+}
+
 function useGetEServiceConsumerActions(
   descriptor?: CatalogEServiceDescriptor,
   delegators?: Array<DelegationTenant>,
   isDelegator?: boolean,
   viewLatestVersionTargetId?: string,
-  requesterEserviceAgreement?: {
-    blocksSubscribe: boolean
-    upgrade?: {
-      agreement: Agreement
-      hasMissingAttributes: boolean
-      hasAllCertifiedAttributes: boolean
-    }
-  }
+  requesterEserviceAgreement?: RequesterEserviceAgreement
 ): {
   primaryAction: ActionItemButton | undefined
   secondaryAction: ActionItemButton | undefined
@@ -266,20 +268,20 @@ function useGetEServiceConsumerActions(
       }
     : undefined
 
-  const upgrade = requesterEserviceAgreement?.upgrade
-  const upgradeAction: ActionItemButton | undefined = upgrade
+  const upgradeAgreement = requesterEserviceAgreement?.upgrade
+  const upgradeAction: ActionItemButton | undefined = upgradeAgreement
     ? {
         action: () =>
           openDialog({
             type: 'upgradeAgreementVersion',
-            agreement: upgrade.agreement,
-            hasMissingAttributes: upgrade.hasMissingAttributes,
+            agreement: upgradeAgreement.agreement,
+            hasMissingAttributes: upgradeAgreement.hasMissingAttributes,
           }),
         label: t('tableEServiceCatalog.upgradeToNewVersion'),
         icon: UpdateIcon,
         variant: 'contained',
-        disabled: !upgrade.hasAllCertifiedAttributes,
-        tooltip: !upgrade.hasAllCertifiedAttributes
+        disabled: !upgradeAgreement.hasAllCertifiedAttributes,
+        tooltip: !upgradeAgreement.hasAllCertifiedAttributes
           ? tAgreement('consumerRead.noCertifiedAttributesForUpgradeTooltip')
           : undefined,
       }
--- a/src/hooks/useGetEServiceConsumerActions.ts
+++ b/src/hooks/useGetEServiceConsumerActions.ts
@@ -1,5 +1,6 @@
 import { AgreementMutations } from '@/api/agreement'
 import type {
+  Agreement,
   AgreementState,
   CatalogEServiceDescriptor,
   DelegationTenant,
@@ -16,6 +17,7 @@ import SendIcon from '@mui/icons-material/Send'
 import PendingActionsIcon from '@mui/icons-material/PendingActions'
 import ArticleIcon from '@mui/icons-material/Article'
 import ReplayCircleFilledIcon from '@mui/icons-material/ReplayCircleFilled'
+import UpdateIcon from '@mui/icons-material/Update'
 import noop from 'lodash/noop'
 import { useDialog } from '@/stores'
 import { match } from 'ts-pattern'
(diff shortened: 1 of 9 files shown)


src/pages/ConsumerEServiceDetailsPage/ConsumerEServiceDetails.page.tsx
```diff
@@ -62,8 +68,56 @@ const ConsumerEServiceDetailsPage: React.FC = () => {
     [descriptor?.eservice.descriptors, descriptorId]
   )
 
+  const consumerAgreementsQuery = useQuery({
+    ...AgreementQueries.getConsumerAgreementsList({
+      limit: 50,
+      offset: 0,
+      eservicesIds: [eserviceId],
+    }),
+    enabled: Boolean(jwt?.organizationId) && !isReviewer,
```
borgesis95: Reviewer can view this page, what happen if this query is not enabled for this role?

src/pages/ConsumerEServiceDetailsPage/ConsumerEServiceDetails.page.tsx
```diff
@@ -62,8 +68,56 @@ const ConsumerEServiceDetailsPage: React.FC = () => {
     [descriptor?.eservice.descriptors, descriptorId]
   )
 
+  const consumerAgreementsQuery = useQuery({
+    ...AgreementQueries.getConsumerAgreementsList({
+      limit: 50,
```
borgesis95: Is necessary put `limit:50`? At most (filtering for `organizationId` and `eserviceId`) should be get only one item.

src/utils/agreement.utils.ts
```diff
@@ -96,6 +97,24 @@ export const canAgreementBeUpgraded = (agreement?: Agreement) => {
   return hasNewVersion && isActiveDescriptorPublishedOrSuspended && isAgreementActiveOrSuspended
 }
 
+export function getRequesterObsoleteVersionAgreement(
+  consumerAgreements: AgreementListEntry[],
+  requesterId: string | undefined
```
borgesis95: What do you think about requesterTenantId? 
```suggestion
  requesterTenantId: string | undefined
```

src/hooks/useGetEServiceConsumerActions.ts
```diff
@@ -253,9 +266,30 @@ function useGetEServiceConsumerActions(
       }
     : undefined
 
+  const upgrade = requesterEserviceAgreement?.upgrade
```
borgesis95:
```suggestion
  const upgradeAgreement = requesterEserviceAgreement?.upgrade
```

src/utils/__tests__/agreement.utils.test.ts
```diff
@@ -331,3 +335,96 @@ describe('getConsumerAgreementVersionAlertSpec utility function testing', () =>
...
+
+describe('getRequesterObsoleteVersionAgreement', () => {
+  it('returns no blocking agreement and no upgradeable id when the list is empty', () => {
+    expect(getRequesterObsoleteVersionAgreement([], 'organizationId')).toEqual({
+      hasBlockingAgreement: false,
+      upgradeableAgreementId: undefined,
+    })
+  })
+
+  it('ignores agreements that belong to other organizations', () => {
```
borgesis95, line 346: This shouldn't be possible (I hope)

src/pages/ConsumerEServiceDetailsPage/ConsumerEServiceDetails.page.tsx
Alepazz: Good point. The query is gated with `enabled: Boolean(jwt?.organizationId) && !isReviewer`, so it never runs for a reviewer. As a result `getRequesterObsoleteVersionAgreement` returns no blocking agreement and `requesterEserviceAgreement` resolves to `undefined`, so a reviewer gets exactly the pre-existing actions, with no blocked/wrong state. I'll add tests for the reviewer (disabled) and the loading branches to make this explicit.

src/pages/ConsumerEServiceDetailsPage/ConsumerEServiceDetails.page.tsx
Alepazz: The endpoint returns the caller's agreements for this e-service, and a consumer can legitimately have more than one over time (historical `ARCHIVED`/`REJECTED` agreements alongside the active/upgradeable one). The helper scans the whole list to find the upgradeable one, so capping at 1 could miss it when it isn't the first item returned. Also `limit` is capped at 50 by the API spec (max 50), so 50 effectively means "all of them", and it's consistent with the other list queries on this page (delegators/delegations). I'd keep it at 50.

src/utils/__tests__/agreement.utils.test.ts
Alepazz: You're right, it shouldn't: `GET /consumers/agreements` is already scoped to the caller, so the list should only ever contain the requester's own agreements (none of the other callers of this endpoint filter by `consumer.id` either). So the `consumer.id === requesterTenantId` filter is a cheap defensive guard rather than a real branch, and this test just documents that guard. I've kept it as insurance, but I'm happy to drop both the filter and this test if you'd rather rely on the endpoint's scoping.

borgesis95 (approved):

PR #2038, opened 2026-06-22
fix: PIN-10400 add line break in archive e-service dialog copy

Alepazz:
## 🔗 Issue

[PIN-10400](https://pagopa.atlassian.net/browse/PIN-10400)

## 📝 Description / Context

In the "Archivia e-service" dialog, the advice paragraph rendered the first sentence's closing period glued to the next sentence ("eliminate.Se confermi"). The source string had no separator at that point, and the host `Typography` did not preserve newlines, so even a newline would have collapsed to a space. The Figma reference shows a line break there, after "…in bozza eliminate." and before "Se confermi questa scelta, …". This adds that line break in both locales and lets the dialog render it.

## 🛠 List of changes

- `it/shared-components.json` and `en/shared-components.json`: added a line break (`\n`) in `dialogArchiveEservice.content.advice.description` after the first sentence (`…eliminate.` / `…deleted.`), to match the Figma reference. The English string mirrors the Italian change for parity.
- `DialogArchiveEservice.tsx`: set `whiteSpace: 'pre-line'` on the advice `Typography` so the newline renders as a line break, consistent with how the project renders `\n` line breaks elsewhere.

## 🧪 How to test

- As a provider admin, open an active e-service and from the three-dots menu choose "Archivia e-service": the first paragraph should break after "…tutte le sue versioni saranno archiviate e quelle in bozza eliminate." with "Se confermi questa scelta, …" starting on a new line.
- Switch the UI to English and repeat: the paragraph should break after "…draft versions will be deleted." with "If you confirm this choice, …" on a new line.

---

PIN-10400-archive-eservice-dialog-en

PIN-10400-archive-eservice-dialog-it

[PIN-10400]: https://pagopa.atlassian.net/browse/PIN-10400

### DIFF
--- a/src/components/dialogs/DialogArchiveEservice.tsx
+++ b/src/components/dialogs/DialogArchiveEservice.tsx
@@ -70,7 +70,7 @@ const DialogArchiveEservice: React.FC<DialogArchiveEserviceProps> = ({ eserviceI
       <FormProvider {...formMethods}>
         <DialogContent>
           {activeStep === 'ADVISE' && (
-            <Typography variant="body2">
+            <Typography variant="body2" sx={{ whiteSpace: 'pre-line' }}>
               <Trans
                 components={{
                   strong: <Typography component="span" variant="inherit" fontWeight={600} />,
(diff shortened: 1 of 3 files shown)


borgesis95 (approved):
Just a quick note about label: In that case is not an `hotfix` (`hotfix` should be used only if something happen in prod and we're fixing it)

PR #2043, opened 2026-06-22
fix: hide "create first purpose" consumer banner when the e-service is being archived (PIN-10429)

Alepazz:
## 🔗 Issue

[PIN-10429](https://pagopa.atlassian.net/browse/PIN-10429)

## 📝 Description / Context

On the consumer "Gestisci richiesta di fruizione" page the blue info banner "Per accedere ai dati crea la tua prima finalità" (a CTA to create the first purpose) was shown whenever the agreement is active and the consumer has no non-archived purposes, with no check on whether the e-service itself is being archived. When the e-service as a whole is being archived a purpose can no longer be created, so the CTA is misleading and must not appear. The single-version case is different: if a version is archived while the e-service stays active (a newer version is published) subscribing is still possible, so the banner must keep showing.

## 🛠 List of changes

- `useGetConsumerAgreementAlertProps`: the `noPurposeAlert` banner is now suppressed when the e-service is being archived, detected via `isDescriptorPendingArchiving(agreement.eservice.activeDescriptor?.state)` (the same signal the provider side already uses for "e-service being archived"). The `suspended` and `missingCertifiedAttributes` alerts are left unchanged.
- Added regression tests for the `ARCHIVING` and `ARCHIVING_SUSPENDED` active-descriptor states (banner hidden), keeping the existing "no purposes" cases (active descriptor `PUBLISHED`) green.

## 🧪 How to test

- As a consumer with an active agreement and no purposes, on an e-service whose active version is being archived (whole e-service archiving): the blue "crea la tua prima finalità" banner is not shown.
- Same setup but only an older version is archiving while a newer version is published (e-service still active): the banner is still shown.
- Unit: `pnpm exec vitest run src/pages/ConsumerAgreementDetailsPage/hooks/__test__/useGetConsumerAgreementAlertProps.test.ts`

[PIN-10429]: https://pagopa.atlassian.net/browse/PIN-10429

### DIFF
--- a/src/pages/ConsumerAgreementDetailsPage/hooks/__test__/useGetConsumerAgreementAlertProps.test.ts
+++ b/src/pages/ConsumerAgreementDetailsPage/hooks/__test__/useGetConsumerAgreementAlertProps.test.ts
@@ -1,4 +1,4 @@
-import type { Agreement, Purpose } from '@/api/api.generatedTypes'
+import type { Agreement, EServiceDescriptorState, Purpose } from '@/api/api.generatedTypes'
 import { renderHookWithApplicationContext } from '@/utils/testing.utils'
 import { useGetConsumerAgreementAlertProps } from '../useGetConsumerAgreementAlertProps'
 import { createMockAgreement } from '../../../../../__mocks__/data/agreement.mocks'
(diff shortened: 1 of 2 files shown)


src/pages/ConsumerAgreementDetailsPage/hooks/useGetConsumerAgreementAlertProps.ts
```diff
@@ -53,7 +54,11 @@ export function useGetConsumerAgreementAlertProps(agreement: Agreement):
     purposes?.results.length === 0 ||
     purposes?.results.every((purpose) => purpose.currentVersion?.state === 'ARCHIVED')
 
-  if (isWithoutPurposes) {
+  const isEServiceBeingArchived = isDescriptorPendingArchiving(
```
borgesis95: Does this also apply when the e-service has been `ARCHIVED` ?

src/pages/ConsumerAgreementDetailsPage/hooks/__test__/useGetConsumerAgreementAlertProps.test.ts
```diff
@@ -167,6 +167,23 @@ describe('check if useGetConsumerAgreementAlertProps returns the correct alertPr
     })
   })
 
+  it.each<EServiceDescriptorState>(['ARCHIVING', 'ARCHIVING_SUSPENDED'])(
+    'shoud not return the noPurposeAlert when the e-service is being archived (active descriptor state %s), even without purposes',
```
LucaPierfederici:
```suggestion
    'should not return the noPurposeAlert when the e-service is being archived (active descriptor state %s), even without purposes',
```
LGTM, just a typo error

LucaPierfederici (approved):

src/pages/ConsumerAgreementDetailsPage/hooks/useGetConsumerAgreementAlertProps.ts
Alepazz: Good catch, it didn't. The previous check only covered the transient archiving states (`ARCHIVING` / `ARCHIVING_SUSPENDED`). I've extended it to also cover the terminal `ARCHIVED` state, so the "create first purpose" banner is now suppressed for a fully archived e-service as well. Added ARCHIVED to the parametrized test too.

borgesis95 (approved):

PR #2046, opened 2026-06-23
fix: show disabled "Richiedi fruizione" button on obsolete e-service version being archived (PIN-10426)

Alepazz:
## 🔗 Issue
[PIN-10426](https://pagopa.atlassian.net/browse/PIN-10426)

## 📝 Description / Context
On the consumer catalog e-service detail page (`Fruizione > Catalogo e-service > Visualizza e-service`), when a non-subscriber viewed an obsolete descriptor that is also being archived (single-descriptor archiving while the e-service stays active because a newer version is published), the disabled "Richiedi fruizione" button was missing entirely. Per the Figma reference ([F07](https://www.figma.com/design/CpRV3kPvFEWLXGtJUgWeZW/Interop-%E2%80%94-Delivery-FE---QA?node-id=8225-70432)) the button must still be shown, but disabled, to signal that fruition cannot be requested on that version. The button was suppressed because the subscribe action is gated by `checkIfcanCreateAgreementDraft`, which returns `false` whenever the viewed descriptor is not `PUBLISHED`/`SUSPENDED` (an obsolete archiving version is in state `ARCHIVING`), and the existing disabled-subscribe fallback only applies while viewing the active descriptor.

## 🛠 List of changes
- `useGetEServiceConsumerActions`: added a `subscribeDisabledArchivingAction` that renders the "Richiedi fruizione" button disabled (no tooltip, matching Figma) when the viewed descriptor is being archived (`isDescriptorPendingArchiving`) and it is an obsolete version (`viewLatestVersionTargetId` set, the same signal that shows the "Vedi ultima versione" link), for a non-subscribed, non-owner, non-delegator admin with no agreement draft. It is appended last in the `primaryAction` fallback chain, so no existing scenario changes.
- Added regression tests for the `ARCHIVING` and `ARCHIVING_SUSPENDED` states (button shown and disabled, no tooltip) plus a negative test confirming no disabled button when the archiving descriptor is the latest version (whole e-service archiving case).

## 🧪 How to test
- As a consumer admin with no agreement on the e-service, open an obsolete version that is being archived while a newer version is published: the "Richiedi fruizione" button is shown but disabled.
- Same page but viewing the latest version: behaviour is unchanged.
- Unit: `pnpm exec vitest run src/hooks/__tests__/useGetEServiceConsumerActions.test.tsx`

---

PIN-10426-disabled-subscribe-obsolete-archiving

[PIN-10426]: https://pagopa.atlassian.net/browse/PIN-10426

### DIFF
--- a/src/hooks/useGetEServiceConsumerActions.ts
+++ b/src/hooks/useGetEServiceConsumerActions.ts
@@ -11,6 +11,7 @@ import {
   checkIfcanCreateAgreementDraft,
   checkIfhasAlreadyAgreementDraft,
 } from '@/utils/agreement.utils'
+import { isDescriptorPendingArchiving } from '@/utils/eservice.utils'
 import { AuthHooks } from '@/api/auth'
 import SendIcon from '@mui/icons-material/Send'
 import PendingActionsIcon from '@mui/icons-material/PendingActions'
@@ -283,6 +284,23 @@ function useGetEServiceConsumerActions(
         }
       : undefined
 
+  const subscribeDisabledArchivingAction: ActionItemButton | undefined =
+    isDescriptorPendingArchiving(descriptor.state) &&
+    Boolean(viewLatestVersionTargetId) &&
+    !isMine &&
+    !isSubscribed &&
+    !hasAgreementDraft &&
+    !isDelegator &&
+    !(delegators && delegators.length > 0)
+      ? {
+          action: noop,
+          label: t('tableEServiceCatalog.subscribe'),
+          icon: SendIcon,
+          variant: 'contained',
+          disabled: true,
+        }
+      : undefined
+
   const handleViewLatestVersion = () => {
     if (viewLatestVersionTargetId) {
       navigate('SUBSCRIBE_CATALOG_VIEW', {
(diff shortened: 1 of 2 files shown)


src/hooks/useGetEServiceConsumerActions.ts
```diff
@@ -283,6 +284,23 @@ function useGetEServiceConsumerActions(
...
       : undefined
 
+  const subscribeDisabledArchivingAction: ActionItemButton | undefined =
+    isDescriptorPendingArchiving(descriptor.state) &&
+    Boolean(viewLatestVersionTargetId) &&
+    !isMine &&
+    !isSubscribed &&
+    !hasAgreementDraft &&
+    !isDelegator &&
+    !(delegators && delegators.length > 0)
```
LucaPierfederici, line 294: One minor change: probably it would be more clean if the condition is extracted in a separate const, as done with `shouldShowhasMissingAttributesTooltip`

LucaPierfederici (approved):

borgesis95 (approved):

src/hooks/useGetEServiceConsumerActions.ts
Alepazz: Done, extracted the condition into a `shouldShowDisabledArchivingSubscribe` const, mirroring `shouldShowhasMissingAttributesTooltip`.

PR #2051, opened 2026-06-23
fix: show disabled subscribe button on obsolete version when the e-service is being archived (PIN-10430)

Alepazz:
## 🔗 Issue

[PIN-10430](https://pagopa.atlassian.net/browse/PIN-10430)

## 📝 Description / Context

On an obsolete (non-latest) e-service version, an eligible non-subscriber should see a disabled, non-clickable "Richiedi fruizione" button while the e-service is being archived. This button was missing: when a newer descriptor exists and the producer starts the e-service archiving, a third-party consumer viewing the old descriptor saw no subscribe action at all.

## 🛠 List of changes

- Add a `subscribeDisabledArchivingAction` in `useGetEServiceConsumerActions` that renders a disabled (no tooltip) subscribe button when the viewed descriptor or the active descriptor is pending archiving (`ARCHIVING` / `ARCHIVING_SUSPENDED`), the user is viewing an obsolete version (`viewLatestVersionTargetId` is set), and is an eligible non-subscriber (not owner, not subscribed, no agreement draft, not a delegator, no delegators).
- Wire it as the last fallback in the `primaryAction` chain, after the enabled subscribe action and the missing-attributes disabled action.
- Add unit tests covering active-descriptor archiving (both `ARCHIVING` and `ARCHIVING_SUSPENDED`), viewed-descriptor archiving, and the negative case (no button when the e-service is not being archived).

## 🧪 How to test

- As a non-subscriber: open an obsolete version of an e-service that has a newer descriptor and is being archived → the grey, non-clickable "Richiedi fruizione" button is shown.
- On an obsolete version of an e-service that is NOT being archived → no subscribe button appears (unchanged behavior).

---

PIN-10430-disabled-subscribe-eservice-archiving

[PIN-10430]: https://pagopa.atlassian.net/browse/PIN-10430


rGregnanin (commented):
Please change the version in the label from 2.21 to 2.22.
Thanks

martinaCampoli (commented):
Now, in the outdated version, the disabled button displays correctly, but if I go on the latest version, which is archiving but still active, the button is not visible, active, or disabled. Is this the correct behaviour?

Image

src/hooks/__tests__/useGetEServiceConsumerActions.test.tsx
```diff
@@ -890,6 +890,101 @@ describe('useGetEServiceConsumerActions tests - actions', () => {
...
+      false,
+      'active-id'
+    )
+
+    expect(result.current.primaryAction?.label).toBe('tableEServiceCatalog.subscribe')
+    expect(result.current.primaryAction?.disabled).toBe(true)
+    expect(result.current.primaryAction?.tooltip).toBeUndefined()
+  })
+
+  it('should not expose the disabled subscribe action on an obsolete version when the e-service is not being archived', () => {
```
alten-dturus, line 959: Maybe I will have add a test that check that the disabled subscribe action is not exposed when I'm in the last version and the eservice is being archived

Alepazz:
> Now, in the outdated version, the disabled button displays correctly, but if I go on the latest version, which is archiving but still active, the button is not visible, active, or disabled. Is this the correct behaviour?
> 
> Image

@martinaCampoli Good point. Right now no subscribe button is shown on the latest version while the e-service is being archived: the enabled subscribe action only applies to PUBLISHED or SUSPENDED versions, and the disabled one introduced by this PR is scoped to obsolete versions only, as per the ticket. The desired behavior for the latest version in this scenario is not specified anywhere, so I would check with Nicolò Cacace, who catalogued this activity, to confirm what should happen in that case. Could you that for me? Thanks

alten-dturus (approved):

antofo (approved):

PR #2053, opened 2026-06-23
fix: show disabled subscribe button on any obsolete e-service version (PIN-10444)

Alepazz:
## 🔗 Issue

[PIN-10444](https://pagopa.atlassian.net/browse/PIN-10444)

## 📝 Description / Context

On an obsolete (non-latest) e-service version, an eligible non-subscriber must see a disabled, non-clickable "Richiedi fruizione" button. [#2051](https://github.com/pagopa/pdnd-interop-frontend/pull/2051) fixed this only while the e-service was being archived; [PIN-10444](https://pagopa.atlassian.net/browse/PIN-10444) reports the same button is also missing on a plain obsolete version that is not being archived. The Figma in the ticket shows the identical disabled button. This PR generalizes the condition so the button appears on any obsolete version, which subsumes the archiving case.

## 🛠 List of changes

- In `useGetEServiceConsumerActions`, drop the archiving-specific guard so the disabled subscribe fallback triggers for an eligible non-subscriber on any obsolete version (`viewLatestVersionTargetId` set, i.e. a newer descriptor exists), regardless of archiving state. Renamed `subscribeDisabledArchivingAction` to `subscribeDisabledObsoleteVersionAction` and removed the now-unused `isDescriptorPendingArchiving` import.
- Consolidated the hook tests into one parametrized positive test covering both the non-archiving obsolete case (PIN-10444) and the archiving obsolete cases (PIN-10430), plus a guard that the button does not appear when viewing the latest version.

## 🧪 How to test

- As a non-subscriber, open an obsolete version of an e-service that has a newer published descriptor and is not being archived → the grey, non-clickable "Richiedi fruizione" button is shown.
- The same holds while the e-service is being archived (unchanged from #2051).
- On the latest version, the button is not shown in its disabled obsolete form (unchanged behavior).

## ⚠️ Known limitation (pre-existing, out of scope)

A delegated consumer (with delegators) can still get an enabled subscribe action on an obsolete version, because `canShowSubscribe` becomes true for multi-tenant delegation flows independently of the descriptor state. This behavior predates this PR and is outside PIN-10444's scope (non-delegated third-party case per the Figma); it is not changed here.

---

PIN-10444-after-disabled-subscribe-obsolete-version

[PIN-10444]: https://pagopa.atlassian.net/browse/PIN-10444

[PIN-10430]: https://pagopa.atlassian.net/browse/PIN-10430?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rGregnanin (commented):
Please change the version in the label from 2.21 to 2.22.
Thanks

rGregnanin (commented):
Please change the version in the label from 2.21 to 2.22.
Thanks

martinaCampoli (approved):

borgesis95 (approved):

PR #2055, opened 2026-06-24
fix: risk analysis assign request snackbar (PIN-10458)

l-giacomini:
## Issue
- [PIN-10458](https://pagopa.atlassian.net/browse/PIN-10458)
- Added snackbar success/error/loading labels for useAssignRiskAnalysisReviewer and useSubmitRiskAnalysis mutations
- Added a parameter to useAssignRiskAnalysisReviewer (it is used in both flows, but only on one of them the snackbar is needed)

[PIN-10458]: https://pagopa.atlassian.net/browse/PIN-10458?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/api/purpose/purpose.mutations.ts
+++ b/src/api/purpose/purpose.mutations.ts
@@ -219,14 +219,16 @@ function useCreateDraftFromPurposeTemplate() {
   })
 }
 
-function useAssignRiskAnalysisReviewer() {
+function useAssignRiskAnalysisReviewer({ showSuccessToast }: { showSuccessToast: boolean }) {
   const { t } = useTranslation('mutations-feedback', {
     keyPrefix: 'purpose.assignRiskAnalysisReviewer',
   })
   return useMutation({
     mutationFn: PurposeServices.assignRiskAnalysisReviewer,
     meta: {
+      successToastLabel: showSuccessToast ? t('outcome.success') : undefined,
       errorToastLabel: t('outcome.error'),
+      loadingLabel: t('loading'),
     },
   })
 }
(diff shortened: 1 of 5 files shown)


borgesis95 (approved):

src/api/purpose/purpose.mutations.ts
```diff
@@ -219,14 +219,16 @@ function useCreateDraftFromPurposeTemplate() {
...
 }
 
-function useAssignRiskAnalysisReviewer() {
+function useAssignRiskAnalysisReviewer({ showSuccessToast }: { showSuccessToast: boolean }) {
   const { t } = useTranslation('mutations-feedback', {
     keyPrefix: 'purpose.assignRiskAnalysisReviewer',
   })
   return useMutation({
     mutationFn: PurposeServices.assignRiskAnalysisReviewer,
     meta: {
```
Alepazz, line 228: This conditional is real logic (`showSuccessToast` decides whether a success toast fires), and `useSubmitRiskAnalysis` now fires one unconditionally, but nothing tests these feedback paths. The component tests mock the whole hook, so they would stay green even if the success toast were always shown, never shown, or wired to the wrong locale key. Could you add a small test asserting that `useAssignRiskAnalysisReviewer({ showSuccessToast: true })` populates `meta.successToastLabel` while `{ showSuccessToast: false }` leaves it undefined? That locks in the behavior this PR adds.

Alepazz (commented):
One note on test coverage below, non blocking.

PR #2056, opened 2026-06-24
fix: sanitize import e-service zip filename suffix (PIN-10161)

GorlemZ:
## 🔗 Issue
[PIN-10161](https://pagopa.atlassian.net/browse/PIN-10161)

## 📝 Description / Context
Importing an e-service via `.zip` fails with `400 Bad Request` (`Invalid zip structure: Error reading configuration.json`, 008-0021) when the external zip file name differs from the archive's internal root folder name.

This happens because the BFF derives the internal root folder from the uploaded file name (`filename` minus `.zip`) instead of reading it from the archive content. The export side couples the two names (`zipFolderName = ${eserviceId}_${descriptorId}` is used both as the inner folder and as the file name), so the import only works while the file is not renamed. As soon as the browser appends a ` (1)` suffix on download (to avoid overwriting an existing file), the invariant breaks and the import is rejected even though `configuration.json` is valid.

This PR is the **frontend mitigation**: it normalizes the uploaded file name before upload so the real production incident (the ` (1)` download-collision suffix) is covered immediately. The **definitive fix lives in the BFF** (read the root folder dynamically from the zip content) and is tracked on the ticket.

## 🛠 List of changes
- Add pure utility `sanitizeImportEserviceFileName` that strips a ` (n)` suffix located right before the `.zip` extension (conservative regex: it leaves legitimate parentheses such as `(copy)` untouched).
- Use the sanitized name in `importVersion` (`eservice.services.ts`) for both the presigned-URL `fileName` param and the `FileResource.filename` sent to the BFF.
- Add data-driven `it.each` tests covering single/multi-digit suffix, untouched names, non-adjacent parentheses, non-`.zip` extensions, and `(copy)`.

## 🧪 How to test
- Export an e-service descriptor and download the resulting `.zip`.
- Rename the local file by adding a ` (1)` suffix before the extension (e.g. `abc_def (1).zip`), or download it twice so the browser adds the suffix automatically.
- Import the file via the e-service import drawer.
- The import succeeds (before this change it failed with `400 Bad Request`).

[PIN-10161]: https://pagopa.atlassian.net/browse/PIN-10161

### DIFF
--- a/src/api/eservice/eservice.services.ts
+++ b/src/api/eservice/eservice.services.ts
@@ -39,6 +39,7 @@ import type {
 } from '../api.generatedTypes'
 import type { AttributeKey } from '@/types/attribute.types'
 import { getAllFromPaginated, waitFor } from '@/utils/common.utils'
+import { sanitizeImportEserviceFileName } from '@/utils/eservice.utils'
 
 async function getCatalogList(params: GetEServicesCatalogParams) {
   const response = await axiosInstance.get<CatalogEServices>(
@@ -450,7 +451,7 @@ async function exportVersion({
 }
 
 async function importVersion({ eserviceFile }: { eserviceFile: File }) {
-  const fileName = eserviceFile.name
+  const fileName = sanitizeImportEserviceFileName(eserviceFile.name)
   const { data: presignedUrl } = await axiosInstance.get<PresignedUrl>(
     `${BACKEND_FOR_FRONTEND_URL}/import/eservices/presignedUrl`,
     { params: { fileName } }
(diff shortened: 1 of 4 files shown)


Alepazz (commented):
One note on test coverage below, non blocking.

src/api/eservice/eservice.services.ts
```diff
@@ -450,7 +451,7 @@ async function exportVersion({
 }
 
 async function importVersion({ eserviceFile }: { eserviceFile: File }) {
```
Alepazz, line 453: The pure helper is well covered, but `importVersion` itself has no test: nothing asserts that the sanitized name is what actually reaches both the presigned-URL `fileName` param and the posted `FileResource.filename`. If one of those two regressed to the raw `eserviceFile.name`, the suite would still pass and this exact bug would silently come back. A service test that mocks `axiosInstance` and checks `toHaveBeenCalledWith` on both calls would lock the fix in. The repo already has this pattern, see `productUpdates.services.test.ts`.

DenisLantero (approved):

Alepazz (approved):

PR #2057, opened 2026-06-24
feature: add accessibility ids to checkbox group component (PIN-2684, INT-57)

alten-dturus:
## 🔗 Issue

[PIN-2684 INT-57](https://pagopa.atlassian.net/browse/A2-2684)

## 📝 Description / Context

The problem in the taks is that `infoLabel` and `error` label does not have an id that refers to the checkbox form component. This PR add the `getAriaAccessibilityInputProps` function to the `RHFCheckboxGroup`, this function returns the `accessibilityProps` and the `ids` that will be passed to the component.

## 🛠 List of changes

- add `getAriaAccessibilityInputProps` to `RHFCheckboxGroup` to generate the `accessibilityProps` and the `ids`
- pass the `accessibilityProps` and the `ids` to the component

### DIFF
--- a/src/components/shared/react-hook-form-inputs/RHFCheckboxGroup.tsx
+++ b/src/components/shared/react-hook-form-inputs/RHFCheckboxGroup.tsx
@@ -43,7 +43,7 @@ export const RHFCheckboxGroup: React.FC<RHFCheckboxGroupProps> = ({
   return (
     <InputWrapper component="fieldset" error={error} sx={sx} infoLabel={infoLabel} {...ids}>
       <FormLabel component="legend">{label}</FormLabel>
-      <FormGroup>
+      <FormGroup {...accessibilityProps}>
         <Controller
           name={name}
           rules={mapValidationErrorMessages(rules, t)}
@@ -72,7 +72,6 @@ export const RHFCheckboxGroup: React.FC<RHFCheckboxGroupProps> = ({
                         checked={field.value?.includes(o.value) ?? false}
                         onChange={onChange}
                         name={String(o.value)}
-                        inputProps={{ ...accessibilityProps }}
                       />
                     }
                     label={o.label}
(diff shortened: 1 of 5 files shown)


src/components/shared/react-hook-form-inputs/RHFCheckboxGroup.tsx
```diff
@@ -35,8 +35,13 @@ export const RHFCheckboxGroup: React.FC<RHFCheckboxGroupProps> = ({
 
   const error = formState.errors[name]?.message as string | undefined
 
+  const { accessibilityProps, ids } = getAriaAccessibilityInputProps(name, {
```
Alepazz, line 38: This PR's whole purpose is the accessibility wiring, but nothing tests it. The current suite only covers toggling, `onValueChange`, and empty options, and it would stay green even if all the ARIA wiring were removed (the `100% coverage on new code` only means these lines run, not that the behavior is asserted). Could you add a test that renders the group with an `infoLabel` and a validation error, then asserts the checkboxes expose `aria-describedby` pointing to the info and error helper ids, plus `aria-invalid` when in error? That locks in the behavior this PR introduces.

src/components/shared/react-hook-form-inputs/RHFCheckboxGroup.tsx
```diff
@@ -67,6 +72,7 @@ export const RHFCheckboxGroup: React.FC<RHFCheckboxGroupProps> = ({
                         checked={field.value?.includes(o.value) ?? false}
                         onChange={onChange}
                         name={String(o.value)}
+                        inputProps={{ ...accessibilityProps }}
```
Alepazz: Optional: `accessibilityProps` is spread onto every checkbox, so with N options each one repeats the same group level description (error plus info) and each gets `aria-invalid` on error. Describing the group once at the fieldset level (the `InputWrapper`) would avoid the repetition. Not a blocker, and it is consistent with how the single `RHFCheckbox` does it, so feel free to leave it as is.

Alepazz (commented):
A couple of notes below, non blocking.

src/components/shared/react-hook-form-inputs/RHFCheckboxGroup.tsx
alten-dturus: I notice that and I think putting the accessibilityProps inside FormGroup can be a valid option

Alepazz (approved):

PR #2068, opened 2026-06-25
fix: handle router thrown errors (PIN-9963)

DenisLantero:
## 🔗 Issue

[PIN-9963](https://pagopa.atlassian.net/browse/PIN-9963)

## 📝 Description / Context

This PR fixes the default React Router fallback shown when a thrown route error is not handled by the application router.

## 🛠 List of changes

- Added a root `errorElement` to the application `createBrowserRouter` configuration.
- Reused the existing `ErrorPage` for errors captured through React Router `useRouteError`.
- Added a regression test to ensure the root route is configured with an `errorElement`.

## 🧪 How to test

- Open an invalid malformed UI route that previously showed React Router's "Unhandled Thrown Error" developer fallback, such as the VAPT scenario from the ticket.
- Verify that the application renders the PDND error page instead of the default React Router developer message.

[PIN-9963]: https://pagopa.atlassian.net/browse/PIN-9963?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/router/components/RoutesWrapper/__tests__/RoutesWrapper.test.tsx
+++ b/src/router/components/RoutesWrapper/__tests__/RoutesWrapper.test.tsx
@@ -1,6 +1,7 @@
 import React from 'react'
 import RoutesWrapper from '../RoutesWrapper'
 import { RouterProvider, createBrowserRouter } from 'react-router-dom'
+import { router } from '@/router/routes'
 import { vi } from 'vitest'
 import * as useTOSAgreement from '../../../hooks/useTOSAgreement'
 import {
@@ -82,6 +83,10 @@ const renderRoutesWrapperWithError = () => {
 }
 
 describe('RoutesWrapper', () => {
+  it('should configure a React Router errorElement on the root route', () => {
+    expect('errorElement' in router.routes[0]).toBe(true)
+  })
+
   it('should show the TOSAgreement when isPublic is false and TOS are not accepted', () => {
     mockUseCurrentRoute({ isPublic: false, routeKey: 'TOS' })
     useTOSAgreementSpy.mockReturnValue({
--- a/src/router/components/RoutesWrapper/__tests__/RoutesWrapper.test.tsx
+++ b/src/router/components/RoutesWrapper/__tests__/RoutesWrapper.test.tsx
@@ -85,7 +85,7 @@ const renderRoutesWrapperWithError = () => {
 const renderRouterErrorPage = () => {
   const rootRoute = router.routes[0]
 
-  if (!('errorElement' in rootRoute)) {
+  if (!('errorElement' in rootRoute) || !React.isValidElement(rootRoute.errorElement)) {
     throw new Error('Expected root route to define an errorElement')
   }
 
--- a/src/router/components/RoutesWrapper/__tests__/RoutesWrapper.test.tsx
+++ b/src/router/components/RoutesWrapper/__tests__/RoutesWrapper.test.tsx
@@ -1,6 +1,6 @@
 import React from 'react'
 import RoutesWrapper from '../RoutesWrapper'
-import { RouterProvider, createBrowserRouter } from 'react-router-dom'
+import { RouterProvider, createBrowserRouter, createMemoryRouter } from 'react-router-dom'
 import { router } from '@/router/routes'
 import { vi } from 'vitest'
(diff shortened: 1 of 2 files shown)


src/router/routes.tsx
```diff
@@ -786,7 +798,11 @@ export type RouteKey = InferRouteKey<typeof routes>
 
 export const router = createBrowserRouter(
   [
-    { element: <RoutesWrapper />, children: reactRouterDOMRoutes },
+    {
+      element: <RoutesWrapper />,
+      errorElement: <RouterErrorPage />,
+      children: reactRouterDOMRoutes,
+    },
     { path: '/', element: <LandingByRole /> },
```
Alepazz: The `errorElement` is attached to only the first of the three sibling top-level routes, so a router-thrown error on `/` or `/*` still falls back to React Router's default error UI. This matters for `/`: `LandingByRole` calls the same `useJwt()` suspense query that can throw `TokenExchangeError`, and unlike the routes under `RoutesWrapper` (wrapped in an `ErrorBoundary`), it has no boundary around it, so a failed token exchange there still surfaces the unhandled-error fallback this PR aims to remove. Wrapping all three under a pathless root route that carries the `errorElement` makes it a real catch-all:

```suggestion
    {
      errorElement: <RouterErrorPage />,
      children: [
        { element: <RoutesWrapper />, children: reactRouterDOMRoutes },
        { path: '/', element: <LandingByRole /> },
        { path: '/*', element: <components.Redirect to="NOT_FOUND" /> },
      ],
    },
```

src/router/routes.tsx
```diff
@@ -92,6 +94,16 @@ const LandingByRole = () => {
   )
 }
 
+const RouterErrorPage = () => {
+  const error = useRouteError()
+
+  return (
+    <Box sx={{ p: 8 }}>
+      <ErrorPage error={error} resetErrorBoundary={() => undefined} />
```
Alepazz: `resetErrorBoundary` is a no-op here, so if an `AxiosError` ever reaches this router-level error page, the retry button rendered by `useResolveError` for that case would do nothing. It is an edge case, since query errors under `RoutesWrapper` are caught by the in-app `ErrorBoundary` where retry actually works, but worth being aware of: a router-level error page may want reload semantics rather than a dead retry.

src/router/components/RoutesWrapper/__tests__/RoutesWrapper.test.tsx
```diff
@@ -81,7 +82,44 @@ const renderRoutesWrapperWithError = () => {
   )
 }
 
+const renderRouterErrorPage = () => {
+  const rootRoute = router.routes[0]
```
Alepazz: This validates that the `errorElement` renders the `ErrorPage` in an isolated memory router, but it does not assert that a route actually left uncovered (for example a throw on `/`, or a malformed URL) is caught in the real `router`. It also couples to `router.routes[0]`, so a change in route order would silently point the helper at the wrong route. A test that drives a sibling-route throw through the real configuration would guard the behavior more directly.

Alepazz (changes requested):
A couple of points before merging, see inline.

src/router/routes.tsx
DenisLantero: Good catch, fixed [here](https://github.com/pagopa/pdnd-interop-frontend/pull/2068/commits/b6dbeb032606906e7f0168da8cd941734f060acf)

src/router/routes.tsx
DenisLantero: Fixed [here](https://github.com/pagopa/pdnd-interop-frontend/pull/2068/commits/a3cca292748e815a80393f6b54be2c5942fb95fd)

src/router/components/RoutesWrapper/__tests__/RoutesWrapper.test.tsx
DenisLantero: Fixed [here](https://github.com/pagopa/pdnd-interop-frontend/pull/2068/commits/dea1a3a8c5d5636bdd8e88b17078a0a72c449c16)

Alepazz (approved):

l-giacomini (approved):

LucaPierfederici (approved):

PR #2070, opened 2026-06-25
fix: avoid polling active queries for notification mark as read (PIN-9620)

DenisLantero:
## 🔗 Issue

[PIN-9620](https://pagopa.atlassian.net/browse/PIN-9620)

## 📝 Description / Context

Opening a provider e-service detail page from "I miei e-service" triggers a silent notification mark-as-read mutation for the current entity.

Before this change, every successful mutation executed the global active-query polling logic in the shared React Query client. This also included silent/background mutations such as marking notifications as read. As a consequence, entering the e-service detail page could start polling and refetch stale active queries that were not strictly related to the current page, producing unnecessary API calls in addition to the expected notification count request.

This change introduces an explicit mutation meta opt-out for cases where a successful mutation should not trigger the global active-query polling. The notification mark-as-read-by-entity mutation now uses this opt-out, keeping the notification behavior unchanged while avoiding unrelated refetches.

## 🛠 List of changes

- Added a `skipActiveQueriesPolling` option to the custom React Query `MutationMeta` type.
- Updated the shared `queryClient` success handler to call active-query polling only when the mutation has not opted out.
- Marked the notification mark-as-read-by-entity mutation with `skipActiveQueriesPolling: true`.
- Added a focused query client test to verify that opt-out mutations do not start the active-query polling interval.
- Preserved the existing global polling behavior for all other mutations.

## 🧪 How to test

- Open the provider area and navigate to "I miei e-service".
- Open the detail page of one e-service.
- Verify that the page still loads correctly and notifications can still be marked as read for the opened entity.
- Verify from the browser network panel that opening the detail page no longer causes unrelated stale active queries to be refetched because of the notification mark-as-read mutation.
- The notification count request may still run as part of the sidebar/header behavior; the important check is that the silent mark-as-read mutation does not trigger additional unrelated polling.

[PIN-9620]: https://pagopa.atlassian.net/browse/PIN-9620?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/api/notification/__tests__/notification.mutations.test.ts
+++ b/src/api/notification/__tests__/notification.mutations.test.ts
@@ -1,0 +1,67 @@
+import { act, waitFor } from '@testing-library/react'
+import { vi } from 'vitest'
+import type * as ReactQuery from '@tanstack/react-query'
+
+const capturedMutationOptions: Array<Record<string, unknown>> = []
+
+vi.mock('@tanstack/react-query', async () => {
+  const actual = await vi.importActual<typeof ReactQuery>('@tanstack/react-query')
+  return {
+    ...actual,
+    useMutation: (options: Record<string, unknown>) => {
+      capturedMutationOptions.push(options)
+      return actual.useMutation(options)
+    },
+  }
+})
+
+const { mockMarkNotificationsAsReadByEntityId } = vi.hoisted(() => ({
+  mockMarkNotificationsAsReadByEntityId: vi.fn(),
+}))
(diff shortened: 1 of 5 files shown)


Alepazz (approved):
LGTM

LucaPierfederici (approved):

PR #2071, opened 2026-06-25
fix: added alert, changed label for RA to be compiled on edit (PIN-10…

l-giacomini:
## Issue
- [PIN-10484](https://pagopa.atlassian.net/browse/PIN-10484)
On the purpose edit page, for a purpose with a risk analysis assigned to a reviewer (to be compiled and approved):
- Changed description label
- Added info alert
- Updated tests

[PIN-10484]: https://pagopa.atlassian.net/browse/PIN-10484?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/static/locales/it/purpose.json
+++ b/src/static/locales/it/purpose.json
@@ -167,10 +167,11 @@
       "readOnlySubtitle": {
         "SUBMITTED": "Non puoi modificare l'analisi del rischio perché è stata inviata in approvazione al valutatore.",
         "SIGNED": "Non è possibile modificare l'analisi del rischio perché è stata approvata dal valutatore.",
-        "ASSIGNED": "Non puoi modificare l'analisi del rischio perché è stata assegnata al valutatore per la compilazione."
+        "ASSIGNED": "Non puoi modificare l’analisi del rischio perché è stata inviata in approvazione al valutatore."
       },
       "rejectedAlert": "Il valutatore ha rifiutato l'analisi del rischio. Modifica l'analisi e inviala di nuovo in approvazione.",
       "rejectedAlertLinkLabel": "Leggi motivazione",
+      "reviewerWritesReviewerSigns": "Quando il valutatore avrà compilato e approvato l’analisi del rischio, potrai pubblicare la finalità.",
       "versionMismatchDialog": {
         "title": "Aggiornamento analisi del rischio",
         "description": "C'è una nuova versione di analisi del rischio disponibile. Quella compilata in precedenza è obsoleta e non può più essere usata. Se clicchi “Aggiorna”, tutti i campi che non sono cambiati ed erano stati compilati li troverai già compilati nella nuova versione",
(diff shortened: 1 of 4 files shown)


borgesis95 (approved):

src/static/locales/it/purpose.json
```diff
@@ -167,10 +167,11 @@
       "readOnlySubtitle": {
         "SUBMITTED": "Non puoi modificare l'analisi del rischio perché è stata inviata in approvazione al valutatore.",
         "SIGNED": "Non è possibile modificare l'analisi del rischio perché è stata approvata dal valutatore.",
-        "ASSIGNED": "Non puoi modificare l'analisi del rischio perché è stata assegnata al valutatore per la compilazione."
+        "ASSIGNED": "Non puoi modificare l’analisi del rischio perché è stata inviata in approvazione al valutatore."
```
Alepazz: I checked the Figma for this screen and the same wording is there, so this matches the design. One thing worth confirming though: the status chip on this card reads `in attesa di compilazione`, while this subtitle says the analysis `è stata inviata in approvazione al valutatore`. Those describe two different states (awaiting compilation vs already submitted for approval), and the new info alert below frames it as still awaiting compilation. Could you double-check with Design that this subtitle is intended as is, given the chip shown next to it?

src/static/locales/it/purpose.json
```diff
@@ -167,10 +167,11 @@
...
       },
       "rejectedAlert": "Il valutatore ha rifiutato l'analisi del rischio. Modifica l'analisi e inviala di nuovo in approvazione.",
       "rejectedAlertLinkLabel": "Leggi motivazione",
+      "reviewerWritesReviewerSigns": "Quando il valutatore avrà compilato e approvato l’analisi del rischio, potrai pubblicare la finalità.",
```
Alepazz, line 170: Small consistency point: the new Italian strings use a curly apostrophe (`’`, as in `l’analisi`), while the nearby `readOnlySubtitle` keys (`SUBMITTED`, `SIGNED`) use a straight one (`l'analisi`). The file mixes both already, so this is minor, but could you either align them here, or open a quick follow-up PR to normalize the apostrophes across the file since it is a fast change?

Alepazz (changes requested):
A couple of points to confirm before merging, see inline.

src/static/locales/it/purpose.json
l-giacomini: @borgesis95 @Alepazz since every translation file mixes these characters, would it be a good idea to open a mixed activity task to normalize them all?
I can start with the purpose.json files right now if you want, but we should decide if we want them like `'` or `’`

src/static/locales/it/purpose.json
l-giacomini: @Alepazz I've forwarded the question to the design team, will let you know as soon as they'll answer!

l-giacomini: @Alepazz The design team along with @borgesis95 have decided to unify these labels in just one, leaving the chips as the state indicator, this is the resulting commit: https://github.com/pagopa/pdnd-interop-frontend/pull/2071/changes/511a9d809b27c4db7c28e1ea0761a96c73c0b22d

Alepazz (approved):

PR #2073, opened 2026-06-25
fix: align readonly assignment step subtitle copy with Figma

LucaPierfederici:
## 🔗 Issue
[PIN-10482](https://pagopa.atlassian.net/browse/PIN-10482)

## 📝 Description / Context
In the "Self-compilation and reviewer approval" flow, after entering "Edit draft" the read-only Assignment step reused the editable form's description ("You can fill in the risk analysis on your own, or entrust a reviewer with the approval or the compilation."). This wrongly suggested the reviewer could still be asked to compile, and did not match the Figma copy.

## 🛠 List of changes
- `PurposeEditStepAssignmentReadOnly` now renders a dedicated subtitle (`readOnly.subtitle`) instead of the shared form `description`
- Added `edit.stepAssignment.readOnly.subtitle` translation key (it / en) with the Figma-approved copy

## 🧪 How to test
- Open a purpose whose risk analysis has been submitted for the reviewer's approval (mode "Self-compilation and reviewer approval")
- Enter the Assignment step in read-only mode (e.g. via "Edit draft")
- Verify the subtitle reads "Non è possibile modificare l'assegnazione perché l'analisi del rischio è stata inviata per l'approvazione al valutatore." (IT) / "The assignment can't be edited because the risk analysis has been submitted to the reviewer for approval." (EN)

[PIN-10482]: https://pagopa.atlassian.net/browse/PIN-10482?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/pages/ConsumerPurposeEditPage/components/PurposeEditStepAssignment/PurposeEditStepAssignmentReadOnly.tsx
+++ b/src/pages/ConsumerPurposeEditPage/components/PurposeEditStepAssignment/PurposeEditStepAssignmentReadOnly.tsx
@@ -35,7 +35,7 @@ const PurposeEditStepAssignmentReadOnly: React.FC<PurposeEditStepAssignmentReadO
 
   return (
     <>
-      <SectionContainer title={t('title')} description={t('description')}>
+      <SectionContainer title={t('title')} description={t('readOnly.subtitle')}>
         <Stack spacing={2}>
           <InformationContainer
             label={t('readOnly.modeLabel')}
--- a/src/static/locales/en/purpose.json
+++ b/src/static/locales/en/purpose.json
@@ -123,6 +123,7 @@
       "forwardBtn": "Save draft and proceed",
       "requestReviewerCompilationBtn": "Request compilation",
       "readOnly": {
+        "subtitle": "The assignment can't be edited because the risk analysis has been submitted to the reviewer for approval.",
         "modeLabel": "Mode",
         "reviewerLabel": "Reviewer",
         "reviewerUnknown": "Reviewer not available",
--- a/src/static/locales/en/purpose.json
+++ b/src/static/locales/en/purpose.json
@@ -123,7 +123,7 @@
       "forwardBtn": "Save draft and proceed",
       "requestReviewerCompilationBtn": "Request compilation",
       "readOnly": {
-        "subtitle": "The assignment can't be edited because the risk analysis has been submitted to the reviewer for approval.",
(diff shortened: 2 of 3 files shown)


borgesis95 (approved):

src/pages/ConsumerPurposeEditPage/components/PurposeEditStepAssignment/PurposeEditStepAssignmentReadOnly.tsx
```diff
@@ -35,7 +35,7 @@ const PurposeEditStepAssignmentReadOnly: React.FC<PurposeEditStepAssignmentReadO
 
   return (
     <>
-      <SectionContainer title={t('title')} description={t('description')}>
+      <SectionContainer title={t('title')} description={t('readOnly.subtitle')}>
```
Alepazz, line 38: The dedicated subtitle is correct for the submitted-for-approval case, but this component is rendered for every read-only entry condition, not just that one. The parent renders it whenever `purpose.reviewerWorkflow` exists or the current version is not an editable draft, so it also covers a published purpose with no reviewer workflow at all (there is a test for that case). In that scenario the subtitle still states the risk analysis was submitted to the reviewer for approval, which is factually wrong since there is no reviewer. It also does not fit the `REVIEWER_WRITES_REVIEWER_SIGNS` mode or the already approved/rejected states. Could you confirm with Design whether this read-only subtitle should vary by state, or whether those other read-only scenarios are out of scope here? Worth noting the new subtitle is not asserted by any test, so if it becomes state-dependent it would be good to cover it.

Alepazz (changes requested):
The dedicated read-only subtitle is a good fix for the submitted-for-approval case, but it is shown for every read-only entry condition, including some where it is inaccurate. One point to address before merging, see inline.

Alepazz (approved):
Approving only to unblock the release given the urgency. Flagging that the main concern from my earlier review is still unaddressed: this subtitle is rendered for every read-only entry condition, so a published purpose with no reviewer workflow still shows it stating the risk analysis was sent to the reviewer, which is inaccurate for that case (there the assignment is read-only simply because the purpose is no longer an editable draft). The latest change only reworded the copy, it did not make the subtitle state-aware. Please follow up to either vary it by state or scope it, so the published purpose with no reviewer case is not misleading.

borgesis95 (approved):

PR #2074, opened 2026-06-25
fix: disabled publish button if purpose has no workflow or ra (PIN-10475)

l-giacomini:
## Issue
- [PIN-10475](https://pagopa.atlassian.net/browse/PIN-10475)
- Added a condition to useGetConsumerPurposesActions.ts for `DRAFT` purpose without a `reviewerWorkflow.signingState` and no riskAnalysisForm compiled to show the delete button only.

[PIN-10475]: https://pagopa.atlassian.net/browse/PIN-10475?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/hooks/useGetConsumerPurposesActions.ts
+++ b/src/hooks/useGetConsumerPurposesActions.ts
@@ -169,8 +169,8 @@ function useGetConsumerPurposesActions(purpose?: Purpose) {
   const hasCurrentVersion = Boolean(purpose?.currentVersion)
   const hasWaitingForApprovalVersion = Boolean(purpose?.waitingForApprovalVersion)
   const hasRejectedVersion = Boolean(purpose?.rejectedVersion)
-  const hasNoReviewerWorkflowAndNoRiskAnalysis =
-    !Boolean(purpose.reviewerWorkflow) && !Boolean(purpose.riskAnalysisForm)
+  const hasReviewerWorkflow = Boolean(purpose.reviewerWorkflow)
+  const hasRiskAnalysisForm = Boolean(purpose.riskAnalysisForm)
 
   const actions = match({
     isDeliverMode,
@@ -185,7 +185,8 @@ function useGetConsumerPurposesActions(purpose?: Purpose) {
     hasCurrentVersion,
     hasWaitingForApprovalVersion,
     hasRejectedVersion,
-    hasNoReviewerWorkflowAndNoRiskAnalysis,
+    hasReviewerWorkflow,
+    hasRiskAnalysisForm,
     routeKey,
   })
     // purpose with no currentVersion but with waitingForApprovalVersion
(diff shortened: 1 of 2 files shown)


src/hooks/useGetConsumerPurposesActions.ts
```diff
@@ -208,6 +211,13 @@ function useGetConsumerPurposesActions(purpose?: Purpose) {
       },
       () => []
     )
+    .with(
+      {
+        isDraft: true,
+        hasNoReviewerWorkflowAndNoRiskAnalysis: true,
+      },
```
Alepazz, line 221: This arm matches any draft with no reviewer workflow and no risk analysis, regardless of mode, and it sits before the generic draft branch. RECEIVE drafts have no risk-analysis step and no assignment step, so they normally have neither `reviewerWorkflow` nor `riskAnalysisForm`, yet they are publishable (the summary page keeps publish enabled for RECEIVE without risk analysis). As written, this would reduce a valid RECEIVE draft to delete only and hide its publish action. Constraining the arm to DELIVER mode keeps RECEIVE drafts on the existing branch:

```suggestion
      {
        isDeliverMode: true,
        isDraft: true,
        hasNoReviewerWorkflowAndNoRiskAnalysis: true,
      },
```

Worth adding a RECEIVE draft test asserting the publish action is still present, since the current test only covers the default mock.

Alepazz (changes requested):
The intent is right for DELIVER, but the new matcher arm also catches RECEIVE drafts, which are publishable without risk analysis. One point to fix before merging, see inline.

src/hooks/useGetConsumerPurposesActions.ts
l-giacomini: Fixed in: https://github.com/pagopa/pdnd-interop-frontend/pull/2074/changes/6871a038b4bc74bb5b632101a0bd80afa79125dc

src/hooks/useGetConsumerPurposesActions.ts
```diff
@@ -208,6 +211,14 @@ function useGetConsumerPurposesActions(purpose?: Purpose) {
       },
       () => []
     )
+    .with(
+      {
+        isDeliverMode: true,
+        isDraft: true,
+        hasNoReviewerWorkflowAndNoRiskAnalysis: true,
```
borgesis95: I'd like to work with this two var separetely

```suggestion
        reviewerWorkflow 
        isAvailableRiskAnalysis
```

l-giacomini: Refactored in: https://github.com/pagopa/pdnd-interop-frontend/pull/2074/changes/9a1529649587e91f075d9c72b823dab1fddf15a4

borgesis95 (approved):

Alepazz (approved):

src/hooks/useGetConsumerPurposesActions.ts
Alepazz: You found the suggest button @borgesis95 😂

PR #2076, opened 2026-06-25
fix: treat whitespace-only values as empty for required field validation (PIN-10488)

Alepazz:
## 🔗 Issue

[PIN-10488](https://pagopa.atlassian.net/browse/PIN-10488)

## 📝 Description / Context

Across the whole platform, required text fields were not validated correctly when filled with only whitespace. react-hook-form's built-in `required` rule treats any non-empty string as filled, so a value like `"   "` was accepted: the field was not flagged in red and the form could be submitted, only to fail later on the backend (the axios request interceptor trims every outgoing string, so the backend received an empty value and returned a validation error). This change makes a whitespace-only value count as empty for the `required` validation, so the field is flagged and blocks submission exactly like a truly empty field.

## 🛠 List of changes

- Added a `withTrimmedRequired` helper in `src/utils/form.utils.ts`: when a `required` rule is present it augments the field rules with a `validate` check that rejects whitespace-only strings with the same `validation.mixed.required` message; it is a no-op for non-string values (e.g. number inputs) and merges with any pre-existing `validate` rule.
- Wired `RHFTextField` to pass its rules through `withTrimmedRequired`. `RHFTextField` is the only free-text input wrapper on the platform, so the shared `mapValidationErrorMessages` and all the other input wrappers (select, checkbox, radio, switch, file, autocomplete) are left untouched.
- Added unit tests for `withTrimmedRequired` and an integration test for `RHFTextField` (typing only spaces into a required field now surfaces the required error).

## ℹ️ Note on fields that also have a minLength constraint

On a required field that also carries a `minLength` (for example 10), react-hook-form evaluates the rules in order and stops at the first failing one, so the message shown while typing only spaces depends on the length: as long as the content stays below 10 characters the "must reach 10 characters" message is shown (the `minLength` rule fails first), and once it reaches 10 or more characters but is still only whitespace the message becomes "Campo richiesto". Both states are blocking, which satisfies the requirement.

## 🧪 How to test

- Open any form with a required text field.
- Leave a required field empty and blur it: it shows "Campo richiesto" in red and blocks submission (unchanged behavior).
- Type only spaces into a required field and blur it: it is now flagged in red with "Campo richiesto" and blocks submission, whereas before it was accepted as filled.
- Type real text with leading/trailing spaces (e.g. `"  hello  "`): no error, only fully-blank content is rejected.
- On a required field that also has a `minLength` (e.g. 10), type only spaces: the "must reach 10 characters" message is shown while under 10 characters, and it becomes "Campo richiesto" once it reaches 10 or more.

## 📸 Screenshots

Screenshot 2026-06-25 alle 16 22 04

Screenshot 2026-06-25 alle 16 22 17

[PIN-10488]: https://pagopa.atlassian.net/browse/PIN-10488


DenisLantero (approved):
LGTM, just fix the CI issues

borgesis95 (approved):

PR #2081, opened 2026-06-29
feat: add staleTime to prefetch queries to avoid duplicate requests (PIN-10389)

Alepazz:
## 🔗 Issue

[PIN-10389](https://pagopa.atlassian.net/browse/PIN-10389)

## 📝 Description / Context

Row-detail tables and catalog cards prefetch the detail data on `onPointerEnter`/`onFocusVisible` via `queryClient.prefetchQuery(...)`. The `QueryClient` sets no default `staleTime`, so it falls back to `0`: every query is immediately stale and `prefetchQuery` re-runs the request on every hover, even when the same entity was just fetched. Hovering across rows or returning to already-visited rows therefore fires many redundant network requests. This PR lets the prefetch serve recently-fetched data from cache within a defined time window.

Design decision: `staleTime` is applied only to the `prefetchQuery` call (by spreading the existing query options), not to the shared `getSingle` query options and not as a global default. Putting it on the shared query options would also make the detail page's own query "fresh" for the window, and the app refreshes data after mutations through a global `mutationCache.onSuccess` that calls `refetchQueries({ type: 'active', stale: true })`, which only refetches stale queries. A fresh detail query would be skipped, so the detail view could show stale data right after a mutation. Scoping `staleTime` to the prefetch keeps the detail observer at the default `staleTime: 0`, so post-mutation polling keeps working as before.

## 🛠 List of changes

- Add shared constant `PREFETCH_STALE_TIME = 60_000` in `src/config/constants.ts`.
- Pass `staleTime: PREFETCH_STALE_TIME` to every hover/focus `queryClient.prefetchQuery(...)` call (24 files, 25 call sites) by spreading the existing query options.
- Leave the one-time session-token prefetch in `App.tsx` untouched (bootstrap call, already `staleTime: Infinity` in its own options).

## 🧪 How to test

- Open the browser DevTools Network tab.
- Go to a list with detail links (e.g. Finalità, Richieste di fruizione, Catalogo e-service).
- Hover repeatedly over the same row action: only the first hover within 60s triggers the `GET`; subsequent hovers on the same entity are served from cache (no new request); a different row still prefetches on first hover; returning to an already-visited row within 60s does not refetch.
- Confirm a detail page still refreshes after a mutation (e.g. suspend/activate): post-mutation refresh must be unaffected.

[PIN-10389]: https://pagopa.atlassian.net/browse/PIN-10389

### DIFF
--- a/src/components/layout/containers/AttributeContainer.tsx
+++ b/src/components/layout/containers/AttributeContainer.tsx
@@ -13,6 +13,7 @@ import { ButtonNaked } from '@pagopa/mui-italia'
 import CheckCircleIcon from '@mui/icons-material/CheckCircle'
 import RemoveCircleOutlineIcon from '@mui/icons-material/RemoveCircleOutline'
 import { AttributeQueries } from '@/api/attribute'
+import { PREFETCH_STALE_TIME } from '@/config/constants'
 import { InformationContainer } from '@pagopa/interop-fe-commons'
 import { useTranslation } from 'react-i18next'
 import { useQuery, useQueryClient } from '@tanstack/react-query'
@@ -80,7 +81,10 @@ export const AttributeContainer = <
   const handlePrefetchAttribute = () => {
     if (alreadyPrefetched.current) return
     alreadyPrefetched.current = true
-    queryClient.prefetchQuery(AttributeQueries.getSingle(attribute.id))
+    queryClient.prefetchQuery({
+      ...AttributeQueries.getSingle(attribute.id),
+      staleTime: PREFETCH_STALE_TIME,
+    })
   }
 
   const isAttributeCertifiedDiscrete =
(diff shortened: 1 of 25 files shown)


DenisLantero (approved):
LGTM, just address the SonarQube review

Alepazz:
On the Sonar coverage gate: the uncovered "new code" is just `staleTime: PREFETCH_STALE_TIME` added inside hover/focus prefetch handlers, repeated across 25 sites. Covering it means ~21 new test files whose only assertion is "the handler called `prefetchQuery`", i.e. tautological mock-pinning tests that catch no real regression. Given the change is mechanical and easily verified by hand (Network tab), I'd propose marking the coverage condition as won't-fix for this PR rather than adding low-value tests. @DenisLantero

l-giacomini (approved):

LucaPierfederici (approved):

borgesis95 (approved):

PR #2082, opened 2026-06-29
fix: encode query params in API requests (PIN-9370, PIN-10613)

DenisLantero:
## 🔗 Issue

[PIN-9370](https://pagopa.atlassian.net/browse/PIN-9370)
[PIN-10613](https://pagopa.atlassian.net/browse/PIN-10613)

## 📝 Description / Context

This PR fixes query parameter serialization for API requests. The custom Axios params serializer was building query strings manually, so reserved characters such as `%` and `&` were not encoded and could produce malformed URLs or incorrect filtering behavior.

The fix uses URL encoding for query parameter keys and values. It also keeps a stricter encoding for characters that `encodeURIComponent` normally leaves unescaped, such as `'`, `!`, `(`, `)`, and `*`, so user-provided search text is always treated as parameter content rather than URL syntax.

## 🛠 List of changes

- URL-encode query parameter keys and values in the shared Axios params serializer.
- Keep stricter percent-encoding for additional special characters left unescaped by `encodeURIComponent`.
- Preserve the existing comma-separated format for array query params.
- Add a focused test covering reserved characters and array serialization.

## 🧪 How to test

- Open an e-service search flow and search by name using reserved characters such as `%`, `&`, or `'`; the request URL should contain encoded values and the query should not be split incorrectly.
- Verify list/autocomplete filters that use query params still work with multi-value filters.

[PIN-9370]: https://pagopa.atlassian.net/browse/PIN-9370?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[PIN-10613]: https://pagopa.atlassian.net/browse/PIN-10613?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/config/axios.ts
+++ b/src/config/axios.ts
@@ -39,7 +39,10 @@ const serializeParams = (query: Record<string, unknown>) => {
       .filter(([, value]) => value !== undefined && value !== null)
       .map(([key, value]) =>
         Array.isArray(value)
-          ? `${encodeQueryParam(key)}=${value.map(encodeQueryParam).join(',')}`
+          ? `${encodeQueryParam(key)}=${value
+              .filter((arrayValue) => arrayValue !== undefined && arrayValue !== null)
+              .map(encodeQueryParam)
+              .join(',')}`
           : `${encodeQueryParam(key)}=${encodeQueryParam(value)}`
       )
       .join('&')
--- a/src/config/axios.ts
+++ b/src/config/axios.ts
@@ -24,6 +24,12 @@ const deepTrim = (object: string | Record<string, unknown>) => {
 }
 
 /** This function helps to serialize correctly arrays in url params  */
+const encodeQueryParam = (value: unknown) =>
+  encodeURIComponent(String(value)).replace(
+    /[!'()*]/g,
+    (char) => `%${char.charCodeAt(0).toString(16).toUpperCase()}`
+  )
+
 const serializeParams = (query: Record<string, unknown>) => {
   return (
     Object.entries(query)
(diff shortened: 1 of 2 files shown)


src/config/axios.ts
```diff
@@ -32,7 +38,9 @@ const serializeParams = (query: Record<string, unknown>) => {
       // like 0 or '' are allowed
       .filter(([, value]) => value !== undefined && value !== null)
       .map(([key, value]) =>
-        Array.isArray(value) ? `${key}=${value.join(',')}` : `${key}=${value}`
+        Array.isArray(value)
+          ? `${encodeQueryParam(key)}=${value.map(encodeQueryParam).join(',')}`
```
Alepazz: The top-level filter drops `undefined`/`null` values, but array members aren't filtered the same way. Since each element now goes through `encodeQueryParam(String(value))`, an array that contains a missing value serializes the literal string `undefined`. For example `eserviceIds`, built from an optional `purpose?.eservice.id` in `useGetProviderPurposesActions` where the query is enabled before `purpose` is loaded, previously serialized to `eserviceIds=` and now serializes to `eserviceIds=undefined`, sending a bogus id to the delegation endpoint. Could you filter nullish members inside arrays too, for symmetry with the top-level filter, and add a test case covering it?

Alepazz (commented):
Encoding fix looks correct and matches the ticket. One point on array params worth addressing before merge, see the inline comment.

src/config/axios.ts
DenisLantero: Fixed [here](https://github.com/pagopa/pdnd-interop-frontend/pull/2082/commits/14a642f7b26bd5d1d4cbe32c03be0efc3fb3590a)

Alepazz (approved):

LucaPierfederici (approved):

l-giacomini (approved):

PR #2083, opened 2026-06-29
fix: disable support confirm actions (PIN-8148)

DenisLantero:
## 🔗 Issue

[PIN-8148](https://pagopa.atlassian.net/browse/PIN-8148)

## 📝 Description / Context

Support users must be able to open backoffice drawers and modals, but they must not be allowed to confirm changes from those UI surfaces.

## 🛠 List of changes

- Added a shared support-action guard provided from the authenticated routes wrapper.
- Disabled mutating confirmation buttons in shared drawers and backoffice dialogs when the current user is `support`.
- Covered representative drawer and modal support-disabled states with focused tests.

## 🧪 How to test

- Log in as a support user and open backoffice drawers or modals that perform updates, approvals, removals, archiving, delegation, or tenant/contact changes.
- Verify that the drawer/modal opens normally, while the confirmation or submit CTA is disabled.
- Verify with a non-support user that the same CTA keeps its previous enabled/disabled behavior.

[PIN-8148]: https://pagopa.atlassian.net/browse/PIN-8148?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/dialogs/DialogUpgradeAgreementVersion.tsx
```diff
@@ -167,12 +169,10 @@ export const DialogUpgradeAgreementVersion: React.FC<DialogUpgradeAgreementVersi
           {t('actions.cancelLabel')}
         </Button>
         <Tooltip
-          title={
-            isUpgradeButtonDisabled ? t('actions.upgrade.notAllCheckboxCheckedTooltip') : undefined
-          }
+          title={isUpgradeDisabled ? t('actions.upgrade.notAllCheckboxCheckedTooltip') : undefined}
```
Alepazz: `isUpgradeDisabled` now also covers the support case, but the tooltip is still hardcoded to `notAllCheckboxCheckedTooltip`. A support user who has checked all three boxes still sees the button disabled with a message about checking the boxes, which is not the real reason. Consider showing no tooltip when the button is disabled only because of the support role, or a dedicated message. The same pattern applies to the shared `Drawer`, where `buttonAction.disabledTooltip` is shown to support users even when the disable reason is the role.

src/hooks/useIsActionDisabledBySupport.ts
```diff
@@ -0,0 +1,21 @@
...
+}
+
+export const SupportActionGuardProvider: React.FC<SupportActionGuardProviderProps> = ({
+  isSupport,
+  children,
+}) => {
+  return React.createElement(SupportActionGuardContext.Provider, { value: isSupport }, children)
+}
+
+export function useIsActionDisabledBySupport(disabled?: boolean) {
```
Alepazz, line 17: Most of the new support-disabled branches are not covered by tests: support-mode tests exist only for `DialogConfirmArchivingAction`, `Drawer`, and `RiskAnalysisVersionMismatchDialog`, while the same wiring in the other dialogs stays unverified. Since the mechanism is centralized here, a focused unit test on `useIsActionDisabledBySupport` (both branches: passthrough of the incoming `disabled`, and forcing `true` under the provider) would cover the core logic with high value. One identical render test per dialog is lower value, but worth deciding explicitly.

src/components/dialogs/DialogSelectAgreementConsumer/DialogSelectAgreementConsumer.tsx
```diff
@@ -161,6 +162,7 @@ export const DialogSelectAgreementConsumer: React.FC<DialogSelectAgreementConsum
     .with('create', () => isQueryEnabled && !isLoading && !hasTenantCertifiedAttributes)
     .with(P.union('edit', 'inspect'), () => !selectedConsumerId)
     .exhaustive()
+  const isButtonActionDisabled = useIsActionDisabledBySupport(isButtonActionDisable)
```
Alepazz: `isButtonActionDisabled` differs from the existing `isButtonActionDisable` by a single trailing letter, which is easy to misread. Consider a clearer name, for example `isConfirmDisabled` as used in the other dialogs.

src/router/components/RoutesWrapper/RoutesWrapper.tsx
```diff
@@ -50,26 +51,28 @@ const _RoutesWrapper: React.FC = () => {
...
-                  <Outlet />
-                </_AuthGuard>
-              </React.Suspense>
-            </ErrorBoundary>
-          </AppLayout>
-        </_TOSGuard>
-        <Footer jwt={jwt} />
-        <Dialog />
-      </Stack>
+      <SupportActionGuardProvider isSupport={isSupport}>
```
Alepazz, line 54: This guard is presentation only: it disables the confirm CTAs in the UI, but a support user could still reach the mutations by other means. Is the backend already rejecting write operations for the support role, or is that still to be verified? Worth confirming so we know whether this is defense in depth or the only barrier.

Alepazz (commented):
Nice work, the shared guard keeps the change consistent across all the dialogs and avoids duplicating the role check. A few observations I would like to see resolved before approving, left inline. None of them affect correctness.

src/components/dialogs/DialogUpgradeAgreementVersion.tsx
DenisLantero: I added the custom tooltip [here](https://github.com/pagopa/pdnd-interop-frontend/pull/2083/commits/268faa9cdbea0a315fa74bb293663bec3bae5a8a)

src/hooks/useIsActionDisabledBySupport.ts
DenisLantero: I fixed it [here](https://github.com/pagopa/pdnd-interop-frontend/pull/2083/commits/936733a29a8632b8383a241824de46deb8b2345e).

I also changed the return because the hook should only force true when the support guard is active, otherwise it should preserve the original disabled value from the caller. 

With `disabled || isSupport`, undefined was being normalized to false.

src/components/dialogs/DialogSelectAgreementConsumer/DialogSelectAgreementConsumer.tsx
DenisLantero: Fixed [here](https://github.com/pagopa/pdnd-interop-frontend/pull/2083/commits/dc517886cc557a1f8896d1d6d27a0588767fbc91)

src/router/components/RoutesWrapper/RoutesWrapper.tsx
DenisLantero: I checked, the BFF itself only authenticates UI users and forwards the auth context, but the processes  reject write operations for the support role.

Alepazz (approved):
Great Job!

LucaPierfederici (approved):
LGTM

borgesis95 (approved):

borgesis95 (approved):

PR #2085, opened 2026-06-30
fix: show cancel estimate change request CTA on consumer purpose detail (PIN-9828)

Alepazz:
## 🔗 Issue

[PIN-9828](https://pagopa.atlassian.net/browse/PIN-9828)

## 📝 Description / Context

On the consumer purpose detail page (Fruizione > Finalità inoltrate > detail), when an estimate-change request exceeds the total threshold it goes into "waiting for approval" and the consumer must be able to withdraw it via the "Annulla richiesta" CTA under the "Nuova stima richiesta" card. The CTA was not shown to a normal consumer: its visibility was gated to `purpose.delegation?.delegator.id === jwt?.organizationId`, a guard introduced with the consumer-delegation feature. For a consumer without a delegation that condition is always false, so the CTA never rendered. This PR aligns the card with its sibling "Richiedi cambio stima" card: the CTA is shown to everyone except the delegator of an active delegation (i.e. the normal consumer and the delegate see it, the delegator does not). It also matches the Figma weight for the link.

## 🛠 List of changes

- `ConsumerPurposeDetailsDailyCallsUpdatePlanCard`: invert the visibility condition from `=== jwt?.organizationId` to `!== jwt?.organizationId`, mirroring the sibling "Richiedi cambio stima" card so the cancel CTA is shown to the normal consumer and the delegate, and hidden for the delegator of an active delegation. The early return (admin + pending version + current version) and the `disabled` on suspended state are unchanged.
- Add `fontWeight: 700` to the "Annulla richiesta" link to match the Figma spec and the sibling card.
- Remove the now-misleading inline comment on the visibility condition.
- Add a test file covering the four role cases (normal admin consumer: shown, non-admin: hidden, delegator: hidden, delegate: shown).

## 🧪 How to test

- As a consumer admin, on a purpose with a pending over-threshold estimate-change request (use "Richiedi cambio stima" with a value above the total threshold), open the purpose detail: under "Nuova stima richiesta" the red, bold "Annulla richiesta" link appears with a trash icon, and clicking it withdraws the request.
- With an active consumer delegation: the delegate sees the link, the delegator does not.
- When the purpose current version is suspended, the link is shown but disabled.

---

PIN-9828-cancel-request-cta

[PIN-9828]: https://pagopa.atlassian.net/browse/PIN-9828

### DIFF
--- a/src/pages/ConsumerPurposeDetailsPage/components/PurposeDetailsTab/__tests__/ConsumerPurposeDetailsDailyCallsUpdatePlanCard.test.tsx
+++ b/src/pages/ConsumerPurposeDetailsPage/components/PurposeDetailsTab/__tests__/ConsumerPurposeDetailsDailyCallsUpdatePlanCard.test.tsx
@@ -1,0 +1,79 @@
+import { screen } from '@testing-library/react'
+import { describe, it, expect, vi } from 'vitest'
+import { ConsumerPurposeDetailsDailyCallsUpdatePlanCard } from '../ConsumerPurposeDetailsDailyCallsUpdatePlanCard'
+import { mockUseJwt, renderWithApplicationContext } from '@/utils/testing.utils'
+import { createMockPurpose } from '@/../__mocks__/data/purpose.mocks'
+import type { PurposeVersion } from '@/api/api.generatedTypes'
+
+vi.mock('@/api/purpose', () => ({
+  PurposeMutations: {
+    useDeleteVersion: () => ({ mutate: vi.fn() }),
+  },
+}))
+
+const waitingForApprovalVersion: PurposeVersion = {
+  id: 'waiting-for-approval-version-id',
+  state: 'WAITING_FOR_APPROVAL',
+  createdAt: '2023-02-03T07:59:52.458Z',
+  dailyCalls: 20,
+}
+
+describe('ConsumerPurposeDetailsDailyCallsUpdatePlanCard', () => {
+  it('shows the remove change plan request link for admin consumers with a pending request', () => {
+    mockUseJwt()
+    renderWithApplicationContext(
+      <ConsumerPurposeDetailsDailyCallsUpdatePlanCard
+        purpose={createMockPurpose({ waitingForApprovalVersion })}
+      />,
+      { withReactQueryContext: true }
+    )
+    expect(screen.getByText('removeChangePlanRequestLink.label')).toBeInTheDocument()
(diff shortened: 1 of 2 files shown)


DenisLantero (approved):
LGTM

l-giacomini (approved):

src/pages/ConsumerPurposeDetailsPage/components/PurposeDetailsTab/__tests__/ConsumerPurposeDetailsDailyCallsUpdatePlanCard.test.tsx
```diff
@@ -0,0 +1,79 @@
...
+const waitingForApprovalVersion: PurposeVersion = {
+  id: 'waiting-for-approval-version-id',
+  state: 'WAITING_FOR_APPROVAL',
+  createdAt: '2023-02-03T07:59:52.458Z',
+  dailyCalls: 20,
+}
+
+describe('ConsumerPurposeDetailsDailyCallsUpdatePlanCard', () => {
+  it('shows the remove change plan request link for admin consumers with a pending request', () => {
+    mockUseJwt()
```
LucaPierfederici: Per our testing convention, `mockUseJwt()` should be called once at module scope (right after the imports, before describe), since 3 of the 4 tests share the same default admin user. Only the non-admin case needs a local override — move the shared call up and keep just `mockUseJwt({ isAdmin: false })` inside that test.

src/pages/ConsumerPurposeDetailsPage/components/PurposeDetailsTab/__tests__/ConsumerPurposeDetailsDailyCallsUpdatePlanCard.test.tsx
```diff
@@ -0,0 +1,79 @@
+import { screen } from '@testing-library/react'
```
LucaPierfederici, line 1: The two behaviors this PR introduces aren't tested: the disabled state when the current version is SUSPENDED (link shown but disabled), and the fontWeight: 700 styling. Worth adding a case that renders with a suspended currentVersion and asserts the link is disabled.

LucaPierfederici (changes requested):

LucaPierfederici (approved):

borgesis95 (approved):

PR #2088, opened 2026-06-30
fix: risk analysis accordion (PIN-10471)

l-giacomini:
## Issue
- [PIN-10471](https://pagopa.atlassian.net/browse/PIN-10471)
- Changed risk analysis summary accordion disposition
- Fixed margins

[PIN-10471]: https://pagopa.atlassian.net/browse/PIN-10471?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/shared/RiskAnalysisInfoSummary.tsx
+++ b/src/components/shared/RiskAnalysisInfoSummary.tsx
@@ -64,7 +64,7 @@ export const RiskAnalysisInfoSummary: React.FC<RiskAnalysisInfoSummaryProps> = (
 
       // Plain text: this value comes from a text field
       if (visualType === 'text') {
-        return { question, answer: answerValue[0] }
+        return { question, answer: answerValue[0], questionInfoLabel }
       }
 
       // Multiple options: this value comes from a multiple choice checkbox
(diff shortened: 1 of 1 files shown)


Alepazz (approved):
One note beyond this diff: free-text answers (`visualType === 'text'`) return the item without `questionInfoLabel`, so an N.B. note on a text-type question is never rendered. It isn't part of these changes, but it would be worth adding for consistency.

DenisLantero (approved):
LGTM, just address @Alepazz comment

l-giacomini:
> One note beyond this diff: free-text answers (`visualType === 'text'`) return the item without `questionInfoLabel`, so an N.B. note on a text-type question is never rendered. It isn't part of these changes, but it would be worth adding for consistency.

questionInfoLabel field added in: https://github.com/pagopa/pdnd-interop-frontend/pull/2088/changes/39cfe53ba82c5a323fe82cd7782717e07473e244

PR #2090, opened 2026-06-30
fix: disabled daily calls if user is reviewer (PIN-10457)

l-giacomini:
## Issue
- [PIN-10457](https://pagopa.atlassian.net/browse/PIN-10457)
- Disabled getRemainingDailyCalls service if user has reviewer role

[PIN-10457]: https://pagopa.atlassian.net/browse/PIN-10457?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/pages/ConsumerPurposeSummaryPage/components/ConsumerPurposeSummaryGeneralInformationAccordion.tsx
+++ b/src/pages/ConsumerPurposeSummaryPage/components/ConsumerPurposeSummaryGeneralInformationAccordion.tsx
@@ -7,6 +7,7 @@ import { useTranslation } from 'react-i18next'
 import { PurposeQueries } from '@/api/purpose'
 import { useQuery, useSuspenseQuery } from '@tanstack/react-query'
 import { useGetPurposeInfoAlert } from '@/hooks/useGetPurposeInfoAlert'
+import { AuthHooks } from '@/api/auth'
 
 type ConsumerPurposeSummaryGeneralInformationAccordionProps = {
   purposeId: string
@@ -15,11 +16,13 @@ type ConsumerPurposeSummaryGeneralInformationAccordionProps = {
 export const ConsumerPurposeSummaryGeneralInformationAccordion: React.FC<
   ConsumerPurposeSummaryGeneralInformationAccordionProps
 > = ({ purposeId }) => {
+  const { isReviewer } = AuthHooks.useJwt()
   const { data: purpose } = useSuspenseQuery(PurposeQueries.getSingle(purposeId))
 
-  const { data: remainingDailyCalls } = useQuery(
-    PurposeQueries.getRemainingDailyCalls({ purposeId })
-  )
+  const { data: remainingDailyCalls } = useQuery({
+    ...PurposeQueries.getRemainingDailyCalls({ purposeId }),
+    enabled: !isReviewer,
+  })
 
   const { t } = useTranslation('purpose', { keyPrefix: 'summary.generalInformationSection' })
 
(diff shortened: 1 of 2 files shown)


src/pages/ConsumerPurposeSummaryPage/components/ConsumerPurposeSummaryGeneralInformationAccordion.tsx
```diff
@@ -15,11 +16,13 @@ type ConsumerPurposeSummaryGeneralInformationAccordionProps = {
...
   ConsumerPurposeSummaryGeneralInformationAccordionProps
 > = ({ purposeId }) => {
+  const { isReviewer } = AuthHooks.useJwt()
   const { data: purpose } = useSuspenseQuery(PurposeQueries.getSingle(purposeId))
 
-  const { data: remainingDailyCalls } = useQuery(
-    PurposeQueries.getRemainingDailyCalls({ purposeId })
-  )
+  const { data: remainingDailyCalls } = useQuery({
+    ...PurposeQueries.getRemainingDailyCalls({ purposeId }),
```
Alepazz, line 23: This disables the remaining daily calls query for reviewers on the assumption that the endpoint is not available to them. That no longer matches the backend: as of [PIN-10284](https://github.com/pagopa/interop-be-monorepo/pull/3448) the `remainingDailyCalls` endpoint explicitly authorizes the reviewer role alongside admin, so a reviewer now gets a 200 with the data, not a 403. The ticket's original expectation (403 for reviewer) predates that backend change. With this guard in place, reviewers would stop seeing the remaining thresholds and the derived alert even though the backend is now designed to serve them. Could you confirm the intended behavior? If reviewers should still see this in the UI, the guard can be dropped. If they intentionally should not, it would help to note why here, since it diverges from the current backend permission.

src/pages/ConsumerPurposeSummaryPage/components/ConsumerPurposeSummaryGeneralInformationAccordion.tsx
```diff
@@ -15,11 +16,13 @@ type ConsumerPurposeSummaryGeneralInformationAccordionProps = {
 export const ConsumerPurposeSummaryGeneralInformationAccordion: React.FC<
   ConsumerPurposeSummaryGeneralInformationAccordionProps
 > = ({ purposeId }) => {
```
Alepazz, line 18: Separately from the behavior question above, this new role-dependent branch is not covered by any test. The page suites that render this accordion stub it out, and there's no dedicated test for the component, so the query gating is never exercised. Once the intended behavior is settled, a small test asserting the query runs for the allowed roles and is skipped otherwise would lock it in.

Alepazz (changes requested):
One point to reconcile before merging, see the inline note.

src/pages/ConsumerPurposeSummaryPage/components/ConsumerPurposeSummaryGeneralInformationAccordion.tsx
l-giacomini: @Alepazz this endpoint has been added to the reviewer role on the BFF as a work-around to avoid getting a 403 in the network tab.
The remaining daily calls are not required from the reviewer prospective, we have just decided to avoid showing a 403 error during the release of the feature, leaving the fix for later.
The BFF will also have to revert this permission.

cc: @borgesis95

src/pages/ConsumerPurposeSummaryPage/components/ConsumerPurposeSummaryGeneralInformationAccordion.tsx
l-giacomini: Tests added in: https://github.com/pagopa/pdnd-interop-frontend/pull/2090/changes/b0978cdd851fe62b52ff8b385056f53857e113c7

Alepazz (approved):

DenisLantero (approved):

PR #2091, opened 2026-06-30
fix/PIN-9877: download type button no submitting

emabruz94:
## 🔗 Issue
[PIN-9877](https://pagopa.atlassian.net/browse/PIN-9877)

## 📝 Description / Context
This PR fixes an unintended form submission when clicking the document download action in the upload documents section.

The download control was rendered as a button without an explicit type, so inside a form it could behave as submit. The fix sets it explicitly as a non-submit button, ensuring the click only triggers file download.

## 🛠 List of changes

- Added explicit button type for the download action in the readonly documents list.
- Prevented accidental form submit when users click to download a document.
- Preserved existing download behavior and UI.

## 🧪 How to test
1. Open the e-service creation flow and reach the documents section.
2. Ensure at least one document is visible in readonly list mode.
3. Click the download action on a document.
4. Verify the file download starts correctly.
5. Verify no form submission or page step transition is triggered by that click.

[PIN-9877]: https://pagopa.atlassian.net/browse/PIN-9877?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/pages/ProviderEServiceCreatePage/components/EServiceCreateStepInfoVersion/components/UploadDocumentsSection.tsx
+++ b/src/pages/ProviderEServiceCreatePage/components/EServiceCreateStepInfoVersion/components/UploadDocumentsSection.tsx
@@ -182,6 +182,7 @@ const UploadDocumentsSectionReadonly: React.FC<{
       <IconLink
         fontWeight={600}
         component="button"
+        type="button"
         onClick={() => handleDownloadDocument(doc)}
         endIcon={<DownloadIcon sx={{ ml: 1 }} fontSize="small" />}
       >


src/pages/ProviderEServiceCreatePage/components/EServiceCreateStepInfoVersion/components/UploadDocumentsSection.tsx
```diff
@@ -182,6 +182,7 @@ const UploadDocumentsSectionReadonly: React.FC<{
       <IconLink
         fontWeight={600}
         component="button"
+        type="button"
```
Alepazz, line 185: Good catch on the explicit submit type. Worth locking this in with a regression test: the current readonly test renders this section outside a form and only checks the document label, and the step test mocks the section entirely, so dropping `type="button"` again would stay green. A small test that renders the readonly section inside a form, clicks the download link, and asserts the form is not submitted would prevent the regression.

Alepazz (approved):

DenisLantero (approved):

l-giacomini (approved):

LucaPierfederici (approved):

PR #2097, opened 2026-07-02
feature: add aria-selected to autocomplete component (PIN-2668, INT-99)

alten-dturus:
## 🔗 Issue

[PIN-2668_INT-99](https://pagopa.atlassian.net/browse/A2-2668)

## 📝 Description / Context

Add aria-selected to autocomplete so the state of the option can be accessible

### DIFF
--- a/src/components/shared/react-hook-form-inputs/RHFAutocomplete/_RHFAutocompleteBase.tsx
+++ b/src/components/shared/react-hook-form-inputs/RHFAutocomplete/_RHFAutocompleteBase.tsx
@@ -159,7 +159,7 @@ export function _RHFAutocompleteBase<
                 />
               )
             }}
-            renderOption={(props, value, { inputValue }) => {
+            renderOption={(props, value, { inputValue, selected }) => {
               const label = value.label
               if (!label) return null
 
@@ -167,7 +167,7 @@ export function _RHFAutocompleteBase<
               const parts = parse(label, matches)
 
               return (
-                <li {...props}>
+                <li {...props} aria-selected={selected}>
                   <div>
                     {parts.map((part, i) => (
                       <Typography


src/components/shared/react-hook-form-inputs/RHFAutocomplete/_RHFAutocompleteBase.tsx
```diff
@@ -159,15 +159,15 @@ export function _RHFAutocompleteBase<
...
+            renderOption={(props, value, { inputValue, selected }) => {
               const label = value.label
               if (!label) return null
 
               const matches = match(label, inputValue, { insideWords: true })
               const parts = parse(label, matches)
 
               return (
-                <li {...props}>
+                <li {...props} aria-selected={selected}>
```
Alepazz, line 170: `props` here already carries `aria-selected`: MUI's `getOptionProps` sets it on the option bag you spread onto the `<li>`, and the `selected` value you destructure is derived from that same `props['aria-selected']`. So re-applying `aria-selected={selected}` after the spread just reassigns the identical value and does not change the rendered DOM. If the accessibility audit flagged a missing `aria-selected` on the options, it was very likely already present, so this may not resolve it. Could you confirm the audit against the actual rendered option element?

Alepazz (commented):
One doubt before merging, see the inline comment. The change looks like a no-op: MUI already sets aria-selected on the option via the spread props, so this likely doesn't move the accessibility audit. Worth double-checking against the rendered DOM.

src/components/shared/react-hook-form-inputs/RHFAutocomplete/_RHFAutocompleteBase.tsx
alten-dturus: I actually tried and the problem is the same: when an option is not selected it is announced while when it is selected it is not. Any ideas?

alten-dturus:
This PR is closed because the mui component is broken, we will use the mui italia component that is accessible to fix

PR #2098, opened 2026-07-02
feature: fix tooltip not readable from all screen readers (PIN-2665, INT-96)

alten-dturus:
## 🔗 Issue

[PIN-2665_INT-96](https://pagopa.atlassian.net/browse/A2-2665)

## 📝 Description / Context

NVDA screen reader doesn't announce the tooltip on eservice create summary pusblish button. This PR include a refactor of the button so the tooltip will be visible to all screen readers

### DIFF
--- a/src/pages/ProviderEServiceSummaryPage/ProviderEServiceSummary.page.tsx
+++ b/src/pages/ProviderEServiceSummaryPage/ProviderEServiceSummary.page.tsx
@@ -513,18 +513,11 @@ const ProviderEServiceSummaryPage: React.FC = () => {
             >
               {tCommon('reject')}
             </Button>
-            <Tooltip title={publishTooltipTitle} arrow>
-              <span>
-                <Button
-                  startIcon={<PublishIcon />}
-                  variant="contained"
-                  onClick={handleApproveDelegatedVersionDraft}
-                  disabled={isSupport || isAsyncExchangeValidationLoading || !isPublishable}
-                >
-                  {tCommon('publish')}
-                </Button>
-              </span>
-            </Tooltip>
+            <PublishButton
+              onClick={handleApproveDelegatedVersionDraft}
+              disabled={isSupport || isAsyncExchangeValidationLoading || !isPublishable}
+              tooltipTitle={publishTooltipTitle}
+            />
           </Stack>
         )}
         {requireDelegateCorrections && sortedRejectedReasons && (
@@ -571,7 +564,7 @@ const PublishButton: React.FC<PublishButtonProps> = ({ disabled, onClick, toolti
           tabIndex={0}
           role="button"
           aria-disabled="true"
-          aria-label={tooltipTitle ? `${buttonLabel} - ${tooltipTitle}` : `${buttonLabel}`}
+          aria-label={tooltipTitle ? `${buttonLabel} - ${tooltipTitle}` : buttonLabel}
         >
           <span aria-hidden="true">{button}</span>
         </span>
--- a/src/pages/ProviderEServiceSummaryPage/ProviderEServiceSummary.page.tsx
+++ b/src/pages/ProviderEServiceSummaryPage/ProviderEServiceSummary.page.tsx
@@ -556,22 +556,30 @@ type PublishButtonProps = {
 
 const PublishButton: React.FC<PublishButtonProps> = ({ disabled, onClick, tooltipTitle }) => {
   const { t: tCommon } = useTranslation('common', { keyPrefix: 'actions' })
+  const buttonLabel = tCommon('publish')
 
-  const Wrapper = disabled
-    ? ({ children }: { children: React.ReactElement }) => (
-        <Tooltip arrow title={tooltipTitle}>
-          <span tabIndex={disabled ? 0 : undefined}>{children}</span>
-        </Tooltip>
-      )
-    : React.Fragment
-
-  return (
-    <Wrapper>
-      <Button disabled={disabled} startIcon={<PublishIcon />} variant="contained" onClick={onClick}>
-        {tCommon('publish')}
-      </Button>
-    </Wrapper>
+  const button = (
+    <Button disabled={disabled} startIcon={<PublishIcon />} variant="contained" onClick={onClick}>
+      {buttonLabel}
+    </Button>
   )
+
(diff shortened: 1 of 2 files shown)


src/pages/ProviderEServiceSummaryPage/ProviderEServiceSummary.page.tsx
```diff
@@ -556,22 +556,30 @@ type PublishButtonProps = {
...
+  if (disabled) {
+    return (
+      <Tooltip arrow title={tooltipTitle}>
+        <span
+          tabIndex={0}
+          role="button"
+          aria-disabled="true"
+          aria-label={`${buttonLabel} - ${tooltipTitle}`}
+        >
+          <span aria-hidden="true">{button}</span>
```
Alepazz, line 569: Hiding the real `<Button disabled>` with `aria-hidden` removes it from the accessibility tree, and the only element left with `role="button"` is this wrapper span, whose accessible name is `publish - <tooltip>` from its `aria-label`. So the existing page tests break: `getByRole('button', { name: 'publish' })` no longer matches (name mismatch, and the native button is hidden), and `toBeDisabled()` does not treat `aria-disabled` on a span as disabled. Please update the disabled publish assertions to the new contract, for example select the wrapper by role and its aria-label and assert `toHaveAttribute('aria-disabled', 'true')`.

src/pages/ProviderEServiceSummaryPage/ProviderEServiceSummary.page.tsx
```diff
@@ -556,22 +556,30 @@ type PublishButtonProps = {
...
   )
+
+  if (disabled) {
+    return (
+      <Tooltip arrow title={tooltipTitle}>
+        <span
+          tabIndex={0}
+          role="button"
+          aria-disabled="true"
+          aria-label={`${buttonLabel} - ${tooltipTitle}`}
```
Alepazz: When the button is disabled but `tooltipTitle` is empty (the `isSupport` path, or while async validation is loading, where the e-service is still publishable so no tooltip key is set), this renders `aria-label="publish - "` with a dangling separator and no reason. Consider adding the separator only when a reason is present:

```suggestion
          aria-label={tooltipTitle ? `${buttonLabel} - ${tooltipTitle}` : buttonLabel}
```

Alepazz (changes requested):
The disabled publish button changes its accessible contract, so the existing page tests that look up the disabled publish action by role button named "publish" and assert toBeDisabled() no longer hold. They need updating to the new markup: query the wrapper by role and aria-label, and assert aria-disabled. See inline.

src/pages/ProviderEServiceSummaryPage/ProviderEServiceSummary.page.tsx
```diff
@@ -556,22 +556,30 @@ type PublishButtonProps = {
...
+  const button = (
+    <Button disabled={disabled} startIcon={<PublishIcon />} variant="contained" onClick={onClick}>
+      {buttonLabel}
+    </Button>
   )
+
+  if (disabled) {
+    return (
+      <Tooltip arrow title={tooltipTitle}>
+        <span
```
Alepazz, line 563: The same accessibility gap remains on the delegated approval publish button (the `isDelegator && descriptor?.state === 'WAITING_FOR_APPROVAL'` branch). It keeps the pattern this PR is removing: a native `disabled` button inside a plain `<span>` wrapped in a `Tooltip`, sharing the same `publishTooltipTitle` and disabled predicates as this CTA, but its `<span>` has no `tabIndex`, `role`, or accessible label. So NVDA and other screen readers still can't reach or read the disabled reason in that flow. Could you apply the same treatment there, ideally by reusing the `PublishButton` component?

src/pages/ProviderEServiceSummaryPage/ProviderEServiceSummary.page.tsx
```diff
@@ -556,22 +556,30 @@ type PublishButtonProps = {
...
+          tabIndex={0}
+          role="button"
+          aria-disabled="true"
+          aria-label={tooltipTitle ? `${buttonLabel} - ${tooltipTitle}` : `${buttonLabel}`}
```
Alepazz: Minor: the else branch wraps a single variable in a template literal. `: buttonLabel` is equivalent and cleaner.
```suggestion
          aria-label={tooltipTitle ? `${buttonLabel} - ${tooltipTitle}` : buttonLabel}
```

Alepazz (changes requested):
The disabled publish CTA fix looks correct, but the delegated approval flow has the same accessibility gap, see inline.

src/pages/ProviderEServiceSummaryPage/__tests__/ProviderEServiceSummary.page.test.tsx
```diff
@@ -240,8 +240,10 @@ describe('ProviderEServiceSummaryPage', () => {
       withRouterContext: true,
     })
 
-    const publishButton = screen.getByRole('button', { name: 'publish' })
-    expect(publishButton).toBeDisabled()
+    const publishButton = screen.getByRole('button', {
+      name: 'publish - summary.notPublishableTooltip.label',
+    })
+    expect(publishButton).toHaveAttribute('aria-disabled', 'true')
```
Alepazz, line 246: `aria-disabled="true"` is hardcoded in the disabled branch of `PublishButton`, so this assertion proves that branch was taken but no longer proves the control is inert, which is what `toBeDisabled()` covered before. The real `<button disabled>` now sits inside an `aria-hidden` subtree, so `getByRole` cannot reach it, and nothing in the suite would fail if `disabled` stopped being forwarded to the inner `Button`: the CTA would become clickable again with all three disabled tests green. Consider asserting the inner control as well. The same applies to the other two disabled publish assertions in this file, the delegated approval one and the expired ruleset one.

```suggestion
    expect(publishButton).toHaveAttribute('aria-disabled', 'true')
    expect(publishButton.querySelector('button')).toBeDisabled()
```

Alepazz (approved):
Approving. One non blocking note on the updated test assertion, see inline.

martinaCampoli (approved):

borgesis95 (approved):

PR #2103, opened 2026-07-07
chore/PIN-10551: release version validation action w CR

emabruz94:
## 🔗 Issue

[PIN-10551](https://pagopa.atlassian.net/browse/PIN-10551)

## 📝 Description / Context

To streamline and automate the process, it often happens that we end up with pull requests that are mergeable into dev (for example, because the next release is the one that needs to be promoted to QA) and other PRs that instead need to be put on hold. 

Post a Change Request (CR) when the PR release label is missing, invalid, or outside the allowed range for dev.

Current CR rules (in short):

A CR is posted if no label in x.y.z format is found on the PR.
A CR is posted if the label does not exactly match ALLOWED_RELEASE_TOWARDS_DEV_BRANCH.
ALLOWED_RELEASE_TOWARDS_DEV_BRANCH is a github variable stored in Settings > Secrets and variables > Variables

The only non-exact match accepted (apart from "align") is a hotfix on the immediately previous minor within the same major:
- major = allowed major
- minor = allowed minor - 1
- patch > 0

Any other case triggers a PR, including:
different major
newer minor than allowed
older than previous minor
previous minor with patch = 0
exception: skip validation using label "align"

## 🛠 List of changes

- new release-version-validation.yml file in github/workflows

## 🧪 How to test

change title or labels on the PR accordingly

[PIN-10551]: https://pagopa.atlassian.net/browse/PIN-10551?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/release-version-validation.yml
+++ b/.github/workflows/release-version-validation.yml
@@ -19,10 +19,12 @@ env:
 jobs:
   validate-dev:
     name: Validate PR towards dev
+    runs-on: ubuntu-latest
     permissions:
       contents: read
       issues: write
       pull-requests: write
+    steps:
       - name: Check if PR is an align PR
         id: check_align
         env:
@@ -55,12 +57,12 @@ jobs:
         if: steps.check_align.outputs.skip != 'true'
         env:
           PR_LABELS_JSON: ${{ toJson(github.event.pull_request.labels.*.name) }}
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_TOKEN: ${{ github.token }}
+        run: |
           echo "Checking PR labels for release tag..."
           echo "Allowed release: $ALLOWED_RELEASE_TOWARDS_DEV_BRANCH"
 
           COMMENT_MARKER='<!-- release-version-validation -->'
-
           get_warning_comment_id() {
             gh api "repos/${{ github.repository }}/issues/${{ github.event.pull_request.number }}/comments" \
               --paginate \
@@ -82,6 +84,7 @@ jobs:
               gh pr comment "${{ github.event.pull_request.number }}" \
                 --repo "${{ github.repository }}" \
                 --body "$body" >/dev/null || true
+            fi
           }
 
           delete_warning_comment() {
@@ -111,11 +114,18 @@ jobs:
 
           ALLOWED_MAJOR=$(echo "$ALLOWED_RELEASE_TOWARDS_DEV_BRANCH" | cut -d. -f1)
           ALLOWED_MINOR=$(echo "$ALLOWED_RELEASE_TOWARDS_DEV_BRANCH" | cut -d. -f2)
-          ALLOWED_PREVIOUS_MINOR=$((ALLOWED_MINOR - 1))
 
           LABEL_MAJOR=$(echo "$RELEASE" | cut -d. -f1)
           LABEL_MINOR=$(echo "$RELEASE" | cut -d. -f2)
           LABEL_PATCH=$(echo "$RELEASE" | cut -d. -f3)
+
+          PREVIOUS_HOTFIX_ALLOWED=false
+          if [ "$ALLOWED_MINOR" -gt 0 ]; then
+            ALLOWED_PREVIOUS_MINOR=$((ALLOWED_MINOR - 1))
+            if [ "$LABEL_MAJOR" -eq "$ALLOWED_MAJOR" ] && [ "$LABEL_MINOR" -eq "$ALLOWED_PREVIOUS_MINOR" ] && [ "$LABEL_PATCH" -gt 0 ]; then
+              PREVIOUS_HOTFIX_ALLOWED=true
+            fi
+          fi
 
           if [ "$RELEASE" = "$ALLOWED_RELEASE_TOWARDS_DEV_BRANCH" ]; then
             echo "Release label '$RELEASE' matches the allowed release. PR is allowed to merge."
@@ -123,12 +133,12 @@ jobs:
             exit 0
           fi
 
-          if [ "$LABEL_MAJOR" -eq "$ALLOWED_MAJOR" ] && [ "$LABEL_MINOR" -eq "$ALLOWED_PREVIOUS_MINOR" ] && [ "$LABEL_PATCH" -gt 0 ]; then
-            echo "Release label '$RELEASE' is a hotfix for the immediately previous minor release. PR is allowed to merge."
+          if [ "$PREVIOUS_HOTFIX_ALLOWED" = "true" ]; then
+            echo "Release label '$RELEASE' is a valid hotfix for the previous release line. PR is allowed to merge."
             delete_warning_comment || true
             exit 0
           fi
 
           echo "::warning::Release label '$RELEASE' is not within the allowed range for merge towards dev. Allowed release: '$ALLOWED_RELEASE_TOWARDS_DEV_BRANCH'."
-          upsert_warning_comment "$(printf '%b' "$COMMENT_MARKER\n⚠️ **Release version validation warning**\n\nRelease label \`$RELEASE\` is not allowed for merge towards \`dev\`.\nAllowed release: \`$ALLOWED_RELEASE_TOWARDS_DEV_BRANCH\`.\nAccepted exception: only the immediately previous minor release with patch greater than 0 is allowed as a hotfix.")" || true
+          upsert_warning_comment "$(printf '%b' "$COMMENT_MARKER\n⚠️ **Release version validation warning**\n\nRelease label \`$RELEASE\` is not allowed for merge towards \`dev\`.\nAllowed release: \`$ALLOWED_RELEASE_TOWARDS_DEV_BRANCH\`.\nAccepted exception: only the immediately previous minor release in the same major, with patch greater than 0, is allowed as a hotfix.")" || true
           exit 0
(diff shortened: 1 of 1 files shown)


.github/workflows/release-version-validation.yml
```diff
@@ -0,0 +1,108 @@
...
+          ALLOWED_PREVIOUS_MINOR=$((ALLOWED_MINOR - 1))
+
+          LABEL_MAJOR=$(echo "$RELEASE" | cut -d. -f1)
+          LABEL_MINOR=$(echo "$RELEASE" | cut -d. -f2)
+          LABEL_PATCH=$(echo "$RELEASE" | cut -d. -f3)
+
+          if [ "$RELEASE" = "$ALLOWED_RELEASE_TOWARDS_DEV_BRANCH" ]; then
+            echo "Release label '$RELEASE' matches the allowed release. PR is allowed to merge."
+            exit 0
+          fi
```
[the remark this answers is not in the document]
DenisLantero: Have you thought about this? Should we add an helper that runs before every validation that deletes the warning if it has been solved?

.github/workflows/release-version-validation.yml
```diff
@@ -0,0 +1,162 @@
...
+  ALLOWED_RELEASE_TOWARDS_DEV_BRANCH: ${{ vars.ALLOWED_RELEASE_TOWARDS_DEV_BRANCH }}
+
+jobs:
+  validate-dev:
+    name: Validate PR towards dev
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
+    steps:
```
borgesis95, line 26: What about add  `continue-on-error: true `and put `exit 1`? In this way we should have pipeline "warning" and you can't go ahead because bot will block merging

.github/workflows/release-version-validation.yml
```diff
@@ -0,0 +1,162 @@
...
+permissions:
+  contents: read
+
+env:
+  ALLOWED_RELEASE_TOWARDS_DEV_BRANCH: ${{ vars.ALLOWED_RELEASE_TOWARDS_DEV_BRANCH }}
+
+jobs:
+  validate-dev:
+    name: Validate PR towards dev
+    runs-on: ubuntu-latest
```
borgesis95, line 21: Should be worth add  `timeout-minutes` var? For instance: what happened if github api doesn't work?

.github/workflows/release-version-validation.yml
```diff
@@ -0,0 +1,162 @@
...
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+env:
+  ALLOWED_RELEASE_TOWARDS_DEV_BRANCH: ${{ vars.ALLOWED_RELEASE_TOWARDS_DEV_BRANCH }}
+
+jobs:
+  validate-dev:
```
borgesis95, line 18: I'd like to start this validation only if PR is 'open' and not in 'draft' state

.github/workflows/release-version-validation.yml
emabruz94: Totally fair suggestion, I'm not sure it's best for our case though.
We basically wanted to keep the pipeline clean and enforce via bot REQUEST_CHANGES.
From what i've understood until now, with continue-on-error:true + exit:1 we wouldn't get a warning, but we would get a "tolerated failure" ("pipeline error but still green") kinda mixed approach that could create confusion, while enforcement still depends on reviews. So maybe it’s clearer to keep this job successful (exit 0) and use review-based blocking only? What do you think it's best?

.github/workflows/release-version-validation.yml
```diff
@@ -0,0 +1,164 @@
...
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize, labeled, unlabeled, edited, ready_for_review]
+    branches:
+      - 'dev'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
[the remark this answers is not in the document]
borgesis95: Take a look here @emabruz94

.github/workflows/release-version-validation.yml
```diff
@@ -0,0 +1,164 @@
...
+          get_latest_bot_review_info() {
+            local reviews_json
+            if ! reviews_json="$(gh api "repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }}/reviews" \
+              --paginate \
+              --slurp)"; then
+              return 1
+            fi
+
+            echo "$reviews_json" | jq -r 'flatten | map(select(.user.login == "github-actions[bot]" and .submitted_at != null)) | sort_by(.submitted_at) | last | [(.id // ""), (.state // ""), (.body // "")] | @tsv'
+          }
```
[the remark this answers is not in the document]
borgesis95: Take a look here @emabruz94

.github/workflows/release-version-validation.yml
```diff
@@ -0,0 +1,164 @@
...
+            local reviews_json
+            if ! reviews_json="$(gh api "repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }}/reviews" \
+              --paginate \
+              --slurp)"; then
+              echo "::warning::Failed to fetch bot reviews; skipping dismissal."
+              return 0
+            fi
+
+            mapfile -t review_ids < <(echo "$reviews_json" | jq -r 'flatten | map(select(.user.login == "github-actions[bot]" and .state == "CHANGES_REQUESTED" and .id != null)) | sort_by(.submitted_at) | .[].id')
+
```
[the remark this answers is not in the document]
borgesis95: take a look here @emabruz94

.github/workflows/release-version-validation.yml
```diff
@@ -0,0 +1,163 @@
...
+            for review_id in "${review_ids[@]}"; do
+              gh api "repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }}/reviews/$review_id/dismissals" \
+                --method PUT \
+                -f message='Release version validation updated. Dismissing previous blocking review.' >/dev/null \
+                || echo "::warning::Failed to dismiss previous blocking review (id: $review_id)."
+            done
+          }
+
+          if [[ "$PR_TITLE" =~ ^align/(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)$ ]]; then
```
antofo: Does it make sense to add a custom label for this check? A label like "ALIGN" of "SKIP-VALIDATION"?

emabruz94: Good point. On the backend they used to skip validation on title "align/x.y.z", but for us it could make more sense to check labels only

.github/workflows/release-version-validation.yml
```diff
@@ -0,0 +1,162 @@
+name: Release version validation
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize, labeled, unlabeled, edited, ready_for_review]
+    branches:
```
DenisLantero: I think this can leave a stale blocking review. For example, a PR targeting `dev` with an invalid `2.19.0` label gets a `REQUEST_CHANGES`. If it is then retargeted to `release/2.19`, GitHub emits `edited` using the new base, so this `dev` filter prevents the workflow from running and the old review is never dismissed.

Could we remove the branch filter, run validation only when the base is `dev`, and dismiss this workflow’s blocking review when an `edited` PR targets another branch?

antofo (approved):
Once the comment of @DenisLantero is resolved I think the pr can be merged

.github/workflows/release-version-validation.yml
emabruz94: This made me think that every "hotfix" PR could not have dev as target but a release branch. 
I would resolve both your issue and this issue by simply having the validation running on any PR targeting dev OR a "release/**" branch (if release branch check if same major and minor). That should do it, what's your thought on this?

DenisLantero: I'm not sure that's the intended behavior tho.

If we add this validation on PRs that target release branches (hotfix PRs) it would mean that we need to add version labels to those PRs as well, and the label would be checked against the version set for dev, which is not correct.

If we don't use the version set for dev, we can't even compare the version label to the branch name since they aren't aligned (ex: `release/2.5.0` --> `2.22.0`).

Personally I think that we should only do what I suggested previously:

> Could we remove the branch filter, run validation only when the base is dev, and dismiss this workflow’s blocking review when an edited PR targets another branch?

Any thoughts?

cc: @borgesis95

emabruz94: Yeh right those versions aren't aligned.. forgot about that. I get what you mean now. 

I don't really like to always run uselessly the validation on any branch if it's not intended to be run on those branches, but I also think it could be a good solution for now and maybe we can come up with something else at a later time. Gonna work on it right now, thanks

.github/workflows/release-version-validation.yml
```diff
@@ -0,0 +1,172 @@
+name: Release version validation
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize, labeled, unlabeled, edited, ready_for_review]
```
DenisLantero: I think the stale-review case still exists when a PR is converted back to draft. Could we also handle `converted_to_draft` and dismiss the workflow’s blocking review? The job condition will need to allow that cleanup path to run for draft PRs.

DenisLantero (approved):

borgesis95 (approved):

PR #2101, opened 2026-07-07
feature: create certified discrete attribute (PIN-10106)

alten-dturus:
## 🔗 Issue

[PIN-10106](https://pagopa.atlassian.net/browse/PIN-10106)

## 📝 Description / Context

This PR add the necessary changes to `CreateAttributeDrawer` inside the certificator page that permit to create a certified discrete attribute

## 🛠 List of changes

- redesign of `CreateAttributeDrawer`
- add new strings

[PIN-10106]: https://pagopa.atlassian.net/browse/PIN-10106?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2106, opened 2026-07-08
feature: assign certified discrete attribute (PIN-10107)

alten-dturus:
[Merge after #2101]

## 🔗 Issue

[PIN-10107](https://pagopa.atlassian.net/browse/PIN-10107)

## 📝 Description / Context

This PR add the necessary changes to `AssignAttributeDrawer` and `DialogRevokeCertifiedAttribute` inside the certificator page that permit to assign and revoke a certified discrete attribute

## 🛠 List of changes

- implement new assign and revoke certified discrete services and mutations
- redesign of `AssignAttributeDrawer`
- add new revoke service to `DialogRevokeCertifiedAttribute`
- add new strings

[PIN-10107]: https://pagopa.atlassian.net/browse/PIN-10107?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/dialogs/DialogRevokeCertifiedAttribute.tsx
+++ b/src/components/dialogs/DialogRevokeCertifiedAttribute.tsx
@@ -14,6 +14,7 @@ import {
 } from '@mui/material'
 import React from 'react'
 import { Trans, useTranslation } from 'react-i18next'
+import { match } from 'ts-pattern'
 
 export const DialogRevokeCertifiedAttribute: React.FC<DialogRevokeCertifiedAttributeProps> = ({
   attribute,
@@ -41,16 +42,20 @@ export const DialogRevokeCertifiedAttribute: React.FC<DialogRevokeCertifiedAttri
   }
 
   const handleRevoke = () => {
-    if (attribute.kind === 'CERTIFIED') {
-      revokeCertifiedAttribute({ tenantId: attribute.tenantId, attributeId: attribute.attributeId })
-    }
-
-    if (attribute.kind === 'CERTIFIED_DISCRETE') {
-      revokeCertifiedDiscreteAttribute({
-        tenantId: attribute.tenantId,
-        attributeId: attribute.attributeId,
-      })
-    }
+    match(attribute.kind)
+      .with('CERTIFIED', () =>
+        revokeCertifiedAttribute({
+          tenantId: attribute.tenantId,
+          attributeId: attribute.attributeId,
+        })
+      )
+      .with('CERTIFIED_DISCRETE', () =>
+        revokeCertifiedDiscreteAttribute({
+          tenantId: attribute.tenantId,
+          attributeId: attribute.attributeId,
+        })
+      )
+      .run()
 
     closeDialog()
   }
--- a/src/components/dialogs/DialogRevokeCertifiedAttribute.tsx
+++ b/src/components/dialogs/DialogRevokeCertifiedAttribute.tsx
@@ -27,6 +27,8 @@ export const DialogRevokeCertifiedAttribute: React.FC<DialogRevokeCertifiedAttri
   const [isConfirmCheckboxChecked, setIsConfirmCheckboxChecked] = React.useState<boolean>(false)
 
   const { mutate: revokeCertifiedAttribute } = AttributeMutations.useRevokeCertifiedAttribute()
+  const { mutate: revokeCertifiedDiscreteAttribute } =
+    AttributeMutations.useRevokeCertifiedDiscreteAttribute()
 
   const handleCheckBoxChange = () => {
     setIsConfirmCheckboxChecked((prev) => {
@@ -39,7 +41,17 @@ export const DialogRevokeCertifiedAttribute: React.FC<DialogRevokeCertifiedAttri
   }
 
   const handleRevoke = () => {
-    revokeCertifiedAttribute({ tenantId: attribute.tenantId, attributeId: attribute.attributeId })
+    if (attribute.kind === 'CERTIFIED') {
+      revokeCertifiedAttribute({ tenantId: attribute.tenantId, attributeId: attribute.attributeId })
+    }
+
+    if (attribute.kind === 'CERTIFIED_DISCRETE') {
+      revokeCertifiedDiscreteAttribute({
+        tenantId: attribute.tenantId,
+        attributeId: attribute.attributeId,
+      })
+    }
+
     closeDialog()
   }
 
--- a/src/pages/TenantCertifierPage/components/AssignAttributesTab/AssignAttributeDrawer.tsx
+++ b/src/pages/TenantCertifierPage/components/AssignAttributesTab/AssignAttributeDrawer.tsx
@@ -141,10 +141,6 @@ export const AssignAttributeDrawer: React.FC<AssignAttributeDrawerProps> = ({
   const isSelectedAttributeCertifiedDiscrete =
     selectedAttribute && selectedAttribute.kind === 'CERTIFIED_DISCRETE'
 
-  React.useEffect(() => {
-    formMethods.resetField('value')
-  }, [formMethods, selectedAttribute])
-
   return (
     <FormProvider {...formMethods}>
       <Drawer
@@ -163,6 +159,7 @@ export const AssignAttributeDrawer: React.FC<AssignAttributeDrawerProps> = ({
             label={t('form.attributeField.label')}
             size="small"
             onInputChange={(_, value) => setAttributeSearchParam(value)}
+            onValueChange={() => formMethods.resetField('value')}
             sx={{ mb: 0, flex: 1 }}
             options={attributeOptions}
             focusOnMount
--- a/src/pages/TenantCertifierPage/components/AssignAttributesTab/AssignAttributeDrawer.tsx
+++ b/src/pages/TenantCertifierPage/components/AssignAttributesTab/AssignAttributeDrawer.tsx
@@ -20,7 +20,7 @@ type AssignAttributeDrawerProps = {
 type AssignAttributeFormValues = {
   attribute: CompactAttribute
   tenant: CompactTenant
-  threshold: number
+  value: number
 }
 
 export const AssignAttributeDrawer: React.FC<AssignAttributeDrawerProps> = ({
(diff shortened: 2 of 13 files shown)


src/pages/TenantCertifierPage/components/AssignAttributesTab/AssignAttributeDrawer.tsx
```diff
@@ -132,23 +149,33 @@ export const AssignAttributeDrawer: React.FC<AssignAttributeDrawerProps> = ({
...
             sx={{ mb: 0, flex: 1 }}
             options={tenantOptions}
             name="tenant"
           />
+          {selectedAttribute && selectedAttribute.kind === 'CERTIFIED_DISCRETE' && (
+            <RHFTextField
+              id="threshold-value-field"
+              label={t('form.thresholdField.label')}
+              name="threshold"
+              rules={{ required: true, min: 0, max: 1000000000 }}
```
Alepazz: The backend contract for `certifiedDiscreteThreshold` is an `int32` with a minimum of 1, but these rules allow `0` and any decimal, so a value the API will reject can be submitted. The existing threshold field in `ConfigureCertifiedDiscreteAttributeDrawer` already enforces `min: 1` plus an integer check via `Number.isInteger`. Could you align this one?

```suggestion
              rules={{ required: true, min: 1, max: 1000000000 }}
```

Consider also adding the integer validator to match the existing drawer.

src/pages/TenantCertifierPage/components/AssignAttributesTab/AssignAttributeDrawer.tsx
```diff
@@ -64,7 +68,7 @@ export const AssignAttributeDrawer: React.FC<AssignAttributeDrawerProps> = ({
     ...AttributeQueries.getList({
       limit: 50,
       offset: 0,
-      kinds: ['CERTIFIED'],
+      kinds: ['CERTIFIED', 'CERTIFIED_DISCRETE'],
```
Alepazz: Elsewhere in the app the `CERTIFIED_DISCRETE` kind is surfaced only behind `FEATURE_FLAG_ATTRIBUTE_CERTIFIED_DISCRETE` (`AttributeAutocomplete`, `AddAttributesToForm/AttributeGroup`, `AttributeContainer`). Here the assign drawer lists discrete attributes and shows the threshold field regardless of the flag. Is the certifier flow meant to bypass it, or should this be gated for consistency?

src/components/dialogs/DialogRevokeCertifiedAttribute.tsx
```diff
@@ -39,7 +41,17 @@ export const DialogRevokeCertifiedAttribute: React.FC<DialogRevokeCertifiedAttri
   }
 
   const handleRevoke = () => {
-    revokeCertifiedAttribute({ tenantId: attribute.tenantId, attributeId: attribute.attributeId })
+    if (attribute.kind === 'CERTIFIED') {
```
Alepazz: Since `kind` is a discriminated union, the project convention is to branch with `ts-pattern` and an `.exhaustive()` check instead of separate `if` statements (the create drawer already does this with `match(kind)...exhaustive()`). As written, `kind` can also be `DECLARED`/`VERIFIED`, and an unhandled value would silently no-op. The same applies to the `onSubmit` handler in `AssignAttributeDrawer`.

Alepazz (changes requested):
See the inline comments

src/components/dialogs/DialogRevokeCertifiedAttribute.tsx
alten-dturus: exhaustive goes in error because not every case is considered. In create drawer the type is defined by me including only two kinds, in this case it is not so I will use ts-pattern but I have to close it with otherwise(() => {})

src/pages/TenantCertifierPage/components/ManageAttributesTab/CreateAttributeDrawer.tsx
```diff
@@ -48,25 +75,63 @@ export const CreateAttributeDrawer: React.FC<CreateAttributeDrawerProps> = ({
...
+        {FEATURE_FLAG_ATTRIBUTE_CERTIFIED_DISCRETE && (
+          <Stack component="form" noValidate spacing={5} mb={3}>
+            (
+            <Stack spacing={3}>
+              <Typography fontWeight={600} variant="label">
+                {t('form.kindField.label')}
+              </Typography>
+              <RHFRadioGroup name="kind" options={kindOptions} rules={{ required: true }} />
+            </Stack>
+            )
```
LucaPierfederici: This `( )` parenthesis are rendered in UI, I think they were not intended to be shown.

src/components/dialogs/DialogRevokeCertifiedAttribute.tsx
alten-dturus: I choose to use run() because the it is like exhaustive but without the problem that should be every case. Obviously if arrives a kind that is not correct the app crashes

LucaPierfederici (changes requested):
I'd add some tests for new code, so sonarqube code coverage let the pipeline pass

borgesis95 (changes requested):
As in the other PR, add tests please

src/pages/TenantCertifierPage/components/AssignAttributesTab/AssignAttributeDrawer.tsx
```diff
@@ -132,23 +154,39 @@ export const AssignAttributeDrawer: React.FC<AssignAttributeDrawerProps> = ({
...
             label={t('form.tenantField.label')}
-            labelType="external"
             size="small"
             onInputChange={(_, value) => setTenantSearchParam(value)}
             sx={{ mb: 0, flex: 1 }}
             options={tenantOptions}
             name="tenant"
           />
+          {selectedAttribute && selectedAttribute.kind === 'CERTIFIED_DISCRETE' && (
+            <RHFTextField
```
Alepazz, line 180: The threshold field is conditionally rendered, but react-hook-form keeps the value of unmounted fields by default (`shouldUnregister` is false). If a user selects a discrete attribute, enters a threshold, switches to a standard certified attribute, then picks another discrete attribute, the field remounts pre-filled with the previous value, so a stale threshold can be submitted for the new attribute. Consider resetting `threshold` (e.g. `resetField('threshold')`) when the selected attribute changes.

src/pages/TenantCertifierPage/components/AssignAttributesTab/AssignAttributeDrawer.tsx
```diff
@@ -132,23 +154,39 @@ export const AssignAttributeDrawer: React.FC<AssignAttributeDrawerProps> = ({
...
           <RHFAutocompleteSingle
             label={t('form.attributeField.label')}
-            labelType="external"
             size="small"
             onInputChange={(_, value) => setAttributeSearchParam(value)}
             sx={{ mb: 0, flex: 1 }}
             options={attributeOptions}
             focusOnMount
             name="attribute"
+            aria-controls="threshold-value-field"
```
Alepazz: `aria-controls` points to `threshold-value-field`, but that input only exists when a discrete attribute is selected, so the reference dangles the rest of the time (including when the flag is off). Consider adding it only when the threshold field is rendered.

src/pages/TenantCertifierPage/components/AssignAttributesTab/__test__/AssignAttributeDrawer.test.tsx
```diff
@@ -0,0 +1,177 @@
...
+      await waitFor(() => {
+        expect(mockAddCertifiedDiscreteAttribute).not.toBeCalled()
+        expect(mockAddCertifiedAttribute).toBeCalledWith({
+          id: 'testAttributeCertifiedId',
+          tenantId: 'test tenant id',
+        })
+      })
+    })
+
+    it('should call addCertifiedDiscreteAttribute if attribute kind selected is CERTIFIED_DISCRETE', async () => {
```
Alepazz, line 226: The tests mock the service functions, so the new POST/DELETE endpoints and payload forwarding are not exercised, and only a valid threshold is covered (the required, min, max, and integer branches are not). Non-blocking, but a case with an invalid threshold plus an assertion on the request the service issues would close the gap the coverage metric misses.

src/static/locales/en/mutations-feedback.json
```diff
@@ -334,11 +334,18 @@
...
+        "success": "You have created a new certified attribute."
+      }
+    },
     "addCertifiedAttribute": {
       "loading": "Attribute is being assigned",
       "outcome": {
-        "error": "The attribute could not be assigned. Please try again",
-        "success": "The attribute has been assigned successfully"
+        "error": "The certificate attribute could not be assigned. Check that you have entered the correct information.",
+        "success": "You have assigned the certified attribute."
```
Alepazz, line 348: This also rewrites the copy of the existing `addCertifiedAttribute` toasts (and the assign drawer subtitle), which affect the plain certified assignment flow, not just the discrete one. Just confirming this copy change is intended for the existing flow.

src/static/locales/en/party.json
```diff
@@ -81,13 +81,23 @@
         "subtitle": "Once created, you can assign and revoke this attribute to other parties",
         "submitBtnLabel": "Create attribute",
         "form": {
-          "nameField": {
-            "label": "Attribute name",
-            "infoLabel": "Min 5 characters, max 160 characters"
+          "kindField": {
+            "label": "Type",
+            "kindRadio": {
+            "optionCertifiedLabel": "Certified attribute",
```
Alepazz, line 87: The `kindRadio` entries are indented one level short here (and in the Italian file). Could you run Prettier on the touched locale files? This block comes from #2101, so it may be cleaner to fix it there.

Alepazz (approved):
The blocking points from the previous review are addressed (threshold validation, the feature-flag gate on the assign drawer, and the ts-pattern dispatch in the revoke dialog). Approving. Left a few non-blocking notes inline, the main one being the threshold value that can carry over when the selected attribute changes.

LucaPierfederici (approved):

src/static/locales/en/mutations-feedback.json
alten-dturus: Yes, it is correct

borgesis95 (approved):

PR #2107, opened 2026-07-08
fix: prevent unchanged attributes update (PIN-10046)

DenisLantero:
## 🔗 Issue

[PIN-10046](https://pagopa.atlassian.net/browse/PIN-10046)

## 📝 Description / Context

This PR fixes an issue in the provider e-service and e-service template details pages where saving the attributes drawer without making any actual change triggered the update mutation and could show a technical error toast.

The chosen behavior is to treat unchanged attributes as a no-op: the drawer closes without calling the backend and without showing a success toast. This keeps the flow aligned with existing multi-field/no-op patterns in the codebase and avoids showing a misleading success notification for an update that was not performed.

## 🛠 List of changes

- Added an unchanged-state guard in `UpdateAttributesDrawer`.
- Prevented e-service and e-service template attribute update mutations when the selected attributes match the initial drawer state.
- Added focused tests for unchanged attributes and for the changed-attributes path.

## 🧪 How to test

- Open a provider e-service details page with existing certified, verified, or declared attributes.
- In the “Thresholds and attributes” section, open the attributes drawer.
- Click “Save changes” without adding any attribute: the drawer should close, no error toast should appear, and no backend update should be triggered.
- Reopen the drawer, add an attribute to an existing group, then save: the update should still be submitted normally.
- Repeat the unchanged-save scenario from an e-service template details page.

[PIN-10046]: https://pagopa.atlassian.net/browse/PIN-10046?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/shared/UpdateAttributesDrawer.tsx
+++ b/src/components/shared/UpdateAttributesDrawer.tsx
@@ -20,6 +20,7 @@ import {
 } from '@/utils/attribute.utils'
 import { useParams } from '@/router'
 import { EServiceTemplateMutations } from '@/api/eserviceTemplate'
+import isEqual from 'lodash/isEqual'
 
 type UpdateAttributesDrawerProps = {
   isOpen: boolean
@@ -103,6 +104,11 @@ export const UpdateAttributesDrawer: React.FC<UpdateAttributesDrawerProps> = ({
   }
 
   const handleSubmit = () => {
+    if (isEqual(selectedAttributes, formModelAttributes)) {
+      onClose()
+      return
+    }
+
     if (kind === 'ESERVICE') {
       updateEserviceAttributes(
         {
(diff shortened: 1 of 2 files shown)


Alepazz (approved):
LGTM. One question: the ticket suggested either blocking the save with a message or closing with a success snackbar, while this PR closes silently without any feedback. Was this third behavior agreed with the Design team?

DenisLantero:
> LGTM. One question: the ticket suggested either blocking the save with a message or closing with a success snackbar, while this PR closes silently without any feedback. Was this third behavior agreed with the Design team?

No I didn't check if this is the wanted behavior with the design team, but I followed the most used pattern for these scenarios.

I believe that adding a toast saying "Success, there were no changes made" doesn't add much to the UX (and it's not what we usually do for these cases), while blocking the save action for the user if no changes were made might be a bit confusing.

If you want I can check with the design team to see if they agree or if they want another solution

Alepazz:
> If you want I can check with the design team to see if they agree or if they want another solution

No need to check that with them, it's okay to me!

l-giacomini (approved):

antofo (approved):

PR #2109, opened 2026-07-08
feat: add generic product updates banner (PIN-9979)

DenisLantero:
## 🔗 Issue

[PIN-9979](https://pagopa.atlassian.net/browse/PIN-9979)

## 📝 Description / Context

This PR updates the product updates banner to support a generic, S3-driven configuration with localized JSON files. The banner content is now read from the current language path, with Italian as fallback, and validated at runtime before being rendered.

## 🛠 List of changes

- Read product updates banner data from `/banner-info-window/{it|en}/data.json`
- Add Italian fallback when the current language file is missing or invalid
- Render dynamic title, description, and up to two optional documentation links
- Add Zod validation for the S3 JSON payload
- Update service, hook, and component tests for localized paths, fallback, validation, and optional links

## 🧪 How to test

- Open the e-service catalog page with language set to Italian and verify the banner uses the `/it/data.json` content.
- Switch to English and verify the banner uses `/en/data.json`; if the English file is unavailable or invalid, verify it falls back to the Italian content.
- Configure `firstLink` and/or `secondLink` in the JSON and verify only configured links are shown and open in a new tab.

[PIN-9979]: https://pagopa.atlassian.net/browse/PIN-9979?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/api/productUpdates/productUpdates.services.ts
+++ b/src/api/productUpdates/productUpdates.services.ts
@@ -1,8 +1,42 @@
 import { APP_MODE, INTEROP_RESOURCES_BASE_URL } from '@/config/env'
 import type { BannerData } from '@/types/banner.types'
 import axios from 'axios'
+import { z } from 'zod'
 
-async function getProductUpdatesJson() {
+const DEFAULT_BANNER_LANGUAGE = 'it'
+
+const productUpdatesBannerLinkSchema = z.object({
+  link: z.string().url(),
+  label: z.string().min(1),
+})
+
+const productUpdatesBannerDataSchema = z.object({
+  start: z.object({
+    date: z.string().min(1),
+    time: z.string().min(1),
+  }),
+  end: z.object({
+    date: z.string().min(1),
+    time: z.string().min(1),
+  }),
+  title: z.string().min(1),
+  description: z.string().min(1),
+  firstLink: productUpdatesBannerLinkSchema.optional(),
+  secondLink: productUpdatesBannerLinkSchema.optional(),
+})
+
+function getBannerLanguage(language: string) {
+  return language.startsWith('en') ? 'en' : DEFAULT_BANNER_LANGUAGE
+}
+
+async function fetchProductUpdatesJson(language: string) {
+  const response = await axios.get<unknown>(
+    `${INTEROP_RESOURCES_BASE_URL}/banner-info-window/${language}/data.json`
+  )
+  return productUpdatesBannerDataSchema.parse(response.data) satisfies BannerData
+}
+
+async function getProductUpdatesJson(language = DEFAULT_BANNER_LANGUAGE) {
   // Mock data in development to avoid CORS issues with S3
   if (APP_MODE === 'development') {
     const bannerData: BannerData = {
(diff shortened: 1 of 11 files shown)


src/api/productUpdates/productUpdates.services.ts
```diff
@@ -1,8 +1,42 @@
 import { APP_MODE, INTEROP_RESOURCES_BASE_URL } from '@/config/env'
 import type { BannerData } from '@/types/banner.types'
 import axios from 'axios'
+import { z } from 'zod'
 
-async function getProductUpdatesJson() {
+const DEFAULT_BANNER_LANGUAGE = 'it'
+
+const productUpdatesBannerLinkSchema = z.object({
+  link: z.string().url(),
```
Alepazz: `z.string().url()` accepts any scheme the `URL` constructor allows, including `javascript:` and `data:`, and the parsed value is passed straight into the anchor `href` of the action buttons. Even if the JSON comes from a trusted S3 source, restricting the schema to http(s) is cheap defense in depth. This also replaces the chained `.url()` form, which is deprecated in the zod version used here, with the top level `z.url()`.

```suggestion
  link: z.url({ protocol: /^https?$/ }),
```

src/components/shared/banners/ProductUpdatesBanner.tsx
```diff
@@ -1,40 +1,48 @@
...
-      // }
+      action1={
+        firstLink && (
+          <Button
+            size="small"
+            color="inherit"
+            endIcon={<OpenInNewIcon />}
+            key="action1"
+            href={firstLink.link}
+            target="_blank"
```
Alepazz, line 26: With `target="_blank"` these links open in a new tab, but their only accessible name is now the `label` coming from the JSON, so screen reader users are no longer told the context changes. The previous implementation signaled this with a dedicated aria label. Consider restoring an accessible "opens in a new tab" hint provided by the frontend (for example via `aria-label`), with the exact wording being a copy decision. This applies to both `action1` and `action2`.

src/hooks/bannerHooks/useProductUpdatesBanner.ts
```diff
@@ -6,8 +6,8 @@ import { useTranslation } from 'react-i18next'
 const STORAGE_KEY = 'productUpdatesBannerDismissedUntil'
 
 export function useProductUpdatesBanner() {
-  const { t } = useTranslation('shared-components', { keyPrefix: 'productUpdatesBanner' })
-  const { data } = useQuery(ProductUpdatesQueries.getProductUpdatesJson())
+  const { i18n } = useTranslation()
```
Alepazz, line 9: Now that the hook reads the banner content from the S3 data instead of the translation files, the `productUpdatesBanner` block (title, body, the action labels and aria labels) in the it and en `shared-components.json` is no longer referenced. It can be removed as cleanup.

Alepazz (changes requested):
A couple of points before merging, see inline comments.

src/api/productUpdates/productUpdates.services.ts
DenisLantero: Good catch, I fixed it [here](https://github.com/pagopa/pdnd-interop-frontend/pull/2109/commits/1be49557d08b6236d407a2b20cabaffc66c03cad)

src/components/shared/banners/ProductUpdatesBanner.tsx
DenisLantero: Added it [here](https://github.com/pagopa/pdnd-interop-frontend/pull/2109/commits/acebb8fbf2251c98a12e1aa0043674db5350d45c)

src/hooks/bannerHooks/useProductUpdatesBanner.ts
DenisLantero: I removed them [here](https://github.com/pagopa/pdnd-interop-frontend/pull/2109/commits/46b3c2b608d9479942f9a3b27545e9d759b4be7c), but I kept the newly added aria label.

martinaCampoli (approved):

Alepazz (approved):
LGTM

PR #2119, opened 2026-07-13
feat: hide facilitated purpose section for receiver eservice (PIN-10582)

antofo:
## Jira Issue
[PIN-10582](https://pagopa.atlassian.net/browse/PIN-10582)

## Context/Why
**Purpose Creation for Receiver EService**

## Key Changes
 - Add check on EService Mode in `PurposeCreateForm` component 

[PIN-10582]: https://pagopa.atlassian.net/browse/PIN-10582?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/pages/ConsumerPurposeCreatePage/components/PurposeCreateForm.tsx
+++ b/src/pages/ConsumerPurposeCreatePage/components/PurposeCreateForm.tsx
@@ -11,7 +11,7 @@ import type {
 import { PurposeMutations, PurposeQueries } from '@/api/purpose'
 import { SectionContainer } from '@/components/layout/containers'
 import { Alert, Box, Button, Divider, Stack } from '@mui/material'
-import React from 'react'
+import React, { useEffect } from 'react'
 import { FormProvider, useForm } from 'react-hook-form'
 import { useTranslation } from 'react-i18next'
 import { PurposeCreateEServiceAutocomplete } from './PurposeCreateEServiceAutocomplete'
@@ -119,7 +119,7 @@ export const PurposeCreateForm: React.FC<PurposeCreateFormProps> = ({ purposeTem
     !isLoadingPurposeTemplates &&
     purposeTemplates?.results &&
     purposeTemplates.results.length > 0 &&
-    selectedEServiceMode !== 'RECEIVE'
+    selectedEServiceMode === 'DELIVER'
 
   // const isSubmitBtnDisabled = !!(useTemplate && purposeId && !purpose)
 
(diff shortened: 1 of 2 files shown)


l-giacomini (approved):

LucaPierfederici (approved):

LucaPierfederici (approved):

src/pages/ConsumerPurposeCreatePage/components/PurposeCreateForm.tsx
```diff
@@ -116,7 +116,10 @@ export const PurposeCreateForm: React.FC<PurposeCreateFormProps> = ({ purposeTem
 
   const selectedEServiceMode = selectedEServiceDescriptor?.eservice.mode
   const isPurposeTemplateCreateSectionVisible =
```
Alepazz, line 118: Hiding the section fixes the direct case, but the invalid flow is still reachable: if the user enables the template on a DELIVER e-service, picks a template, then switches to a RECEIVE one, the section unmounts but `usePurposeTemplate` and `purposeTemplateId` stay in the form (the reset effect lives inside the now-unmounted section). The template branch in `onSubmit` then runs regardless of mode and hits the same 400. Consider guarding it on `selectedEServiceMode === 'DELIVER'`, or clearing the template fields when the mode isn't DELIVER.

src/pages/ConsumerPurposeCreatePage/components/PurposeCreateForm.tsx
```diff
@@ -116,7 +116,10 @@ export const PurposeCreateForm: React.FC<PurposeCreateFormProps> = ({ purposeTem
...
-    !isLoadingPurposeTemplates && purposeTemplates?.results && purposeTemplates.results.length > 0
+    !isLoadingPurposeTemplates &&
+    purposeTemplates?.results &&
+    purposeTemplates.results.length > 0 &&
+    selectedEServiceMode !== 'RECEIVE'
```
Alepazz: Since `EServiceMode` is only `'RECEIVE' | 'DELIVER'`, `!== 'RECEIVE'` also treats the "mode not yet known" state (descriptor still loading or failed to resolve) as visible, so the section briefly flashes when a RECEIVE e-service is selected. A positive `=== 'DELIVER'` keeps it hidden until the mode is confirmed and matches the ticket's "exclusively for DELIVER". Note the existing tests rely on visibility with an undefined mode, so they'd need a DELIVER descriptor mock if you switch.

Alepazz (changes requested):
A couple of points before merging, see inline comments.

src/pages/ConsumerPurposeCreatePage/components/PurposeCreateForm.tsx
antofo: resolved

src/pages/ConsumerPurposeCreatePage/components/PurposeCreateForm.tsx
antofo: resolved

borgesis95 (approved):

PR #2120, opened 2026-07-13
fix: prevent decimal values in eservice thresholds (PIN-9642)

DenisLantero:
## 🔗 Issue

[PIN-9642](https://pagopa.atlassian.net/browse/PIN-9642)

## 📝 Description / Context

Prevent decimal values from being entered in the expected API calls threshold fields during eservice creation. This avoids sending floating-point values to APIs that only accept integers and prevents the resulting generic error.

## 🛠 List of changes

- Added an `integerOnly` mode to the shared React Hook Form numeric field.
- Enabled integer-only input for the per-consumer and total daily calls thresholds.
- Set the input step to `1` for both threshold fields.
- Added regression tests covering decimal input in both fields.

## 🧪 How to test

- Start the eservice creation flow and navigate to the thresholds and attributes step.
- Try entering a decimal value, such as `1.5` or `1,5`, in both “API calls/day per consumer” and “Total API calls/day”.
- Verify that decimal separators cannot be entered.
- Enter valid integer values and verify that the step can be saved successfully.

[PIN-9642]: https://pagopa.atlassian.net/browse/PIN-9642?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/shared/react-hook-form-inputs/RHFTextField.tsx
+++ b/src/components/shared/react-hook-form-inputs/RHFTextField.tsx
@@ -25,10 +25,12 @@ export type RHFTextFieldProps = Omit<MUITextFieldProps, 'type' | 'label'> & {
 } & (
     | {
         type?: 'text' | 'email'
+        integerOnly?: never
         onValueChange?: (value: string) => void
       }
     | {
         type?: 'number'
+        integerOnly?: boolean
         onValueChange?: (value: number) => void
       }
   )
@@ -41,6 +43,7 @@ export const RHFTextField: React.FC<RHFTextFieldProps> = ({
   infoLabel,
   focusOnMount,
   multiline,
+  integerOnly = false,
   onValueChange,
   rules,
   size = 'small',
@@ -104,7 +107,8 @@ export const RHFTextField: React.FC<RHFTextFieldProps> = ({
                 ? (e) => {
                     if (e.key.length !== 1 || e.ctrlKey || e.metaKey) return
                     const isEmpty = (e.target as HTMLInputElement).value === ''
-                    if (!/[1-9.\-]/.test(e.key) && !(e.key === '0' && !isEmpty)) {
+                    const allowedCharacters = integerOnly ? /[1-9\-]/ : /[1-9.\-]/
+                    if (!allowedCharacters.test(e.key) && !(e.key === '0' && !isEmpty)) {
                       e.preventDefault()
                     }
                   }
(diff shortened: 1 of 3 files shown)


src/components/shared/react-hook-form-inputs/RHFTextField.tsx
```diff
@@ -104,7 +107,8 @@ export const RHFTextField: React.FC<RHFTextFieldProps> = ({
                 ? (e) => {
                     if (e.key.length !== 1 || e.ctrlKey || e.metaKey) return
                     const isEmpty = (e.target as HTMLInputElement).value === ''
-                    if (!/[1-9.\-]/.test(e.key) && !(e.key === '0' && !isEmpty)) {
+                    const allowedCharacters = integerOnly ? /[1-9\-]/ : /[1-9.\-]/
```
Alepazz, line 110: Blocking only the separator here means the digits typed after it still get appended to the integer part: typing `1.5` yields `15`, and `100.5` yields `1005`. Since the result is a valid integer, the `Number.isInteger` rule on the threshold fields does not catch it, so no error is shown and a value roughly ten times larger than intended is submitted silently.

Worth noting that today, typing a decimal with the period surfaces the clear "integers only" validation message on these fields; this change replaces that explicit feedback with a silent conversion. Could we either rely on the existing `validate` rule (which already rejects decimals from any input path, including paste, with a clear message), or make the real-time guard not produce a misleading value?

Alepazz (changes requested):
One doubt on the integer-only handling before merging, left inline.

src/components/shared/react-hook-form-inputs/RHFTextField.tsx
DenisLantero: Good catch, you're right: blocking the separator makes `1.5` become `15`, so this change would actually regress the current behavior.

One thing tho, the existing validation already handles `1.5` correctly, but typing `1,5` on `dev` still produces `15`.

How would you approach that case? Would you normalize `,` to `.` before letting the existing integer validation handle it, or do you see a cleaner solution?

DenisLantero: P.S, I've checked and it seems with the current behavior in dev the issue isn't reproducible anymore, but we could still think about normalizing the , to .

antofo (approved):

l-giacomini (approved):

src/components/shared/react-hook-form-inputs/RHFTextField.tsx
Alepazz: @DenisLantero I wouldn’t normalize the comma globally in `RHFTextField`, since it is shared across several flows. Since the original issue no longer seems reproducible on `dev`, I’d first ask QA to retest it there.

If it is still reproducible, both `1.5` and `1,5` should produce the existing integer validation error without being silently transformed into `15`. The regression tests should cover both separators and assert the validation error rather than the converted value. Otherwise, if QA confirms the issue is already fixed on `dev`, I think this PR can be closed.

PR #2129, opened 2026-07-16
feat: move agreement id from body to query (PIN-10630)

LucaPierfederici:
## 🔗 Issue
[PIN-10630](https://pagopa.atlassian.net/browse/PIN-10630)

## 📝 Description / Context
The BFF changed the contract for the verified attribute revocation endpoint: `agreementId` is now expected as a query parameter instead of a request body. `DELETE` requests with a body are unreliable and the new contract drops it entirely, so the frontend must align or the revocation fails.

The regenerated API types also pick up the split of the agreement activation endpoint into `approve` (`POST /agreements/{agreementId}/approve`) and `unsuspend` (`POST /agreements/{agreementId}/unsuspend`), replacing the previous single `activate` endpoint.

## 🛠 List of changes
- `revokeVerifiedPartyAttribute` now sends `agreementId` via `params` instead of `data`, so it travels as a query parameter with no request body
- Regenerated API types: `RevokeVerifiedAttributePayload` removed, `agreementId` moved into `RevokeVerifiedAttributeParams` and the request body is now `never`
- Regenerated API types: `ActivateAgreementParams` split into `ApproveAgreementParams` and `UnsuspendAgreementParams`, with the `ActivateAgreement` namespace replaced by `ApproveAgreement` and `UnsuspendAgreement`
- Added unit test covering that the revocation sends the agreement id as a query parameter without a body

## 🧪 How to test
- Log in as a provider and open the details page of an agreement that has verified attributes
- Open the verified attributes drawer and revoke a verified attribute
- In the network tab, check the `DELETE /tenants/{partyId}/attributes/verified/{attributeId}` request: `agreementId` must appear in the query string and the request must have no body
- Verify the revocation completes successfully and the attribute state updates in the UI

[PIN-10630]: https://pagopa.atlassian.net/browse/PIN-10630?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/api/api.generatedTypes.ts
+++ b/src/api/api.generatedTypes.ts
@@ -3017,7 +3017,15 @@ export interface DeleteAgreementParams {
   agreementId: string;
 }
 
-export interface ActivateAgreementParams {
+export interface ApproveAgreementParams {
+  /**
+   * The identifier of the agreement
+   * @format uuid
+   */
+  agreementId: string;
+}
+
+export interface UnsuspendAgreementParams {
   /**
    * The identifier of the agreement
    * @format uuid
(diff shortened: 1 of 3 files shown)


borgesis95:
This PR is in `DRAFT` state because before to work on it this https://github.com/pagopa/interop-be-monorepo/pull/3609 has to be merged

PR #2136, opened 2026-07-21
feat: add archive eservice dialog guidance (PIN-10659)

DenisLantero:
## 🔗 Issue

[PIN-10659](https://pagopa.atlassian.net/browse/PIN-10659)

## 📝 Description / Context

Add guidance to the e-service archiving dialog to clarify the constraint applied when one of its versions is already being archived.

## 🛠 List of changes

- Add the guidance below the notice period selection in the e-service archiving dialog.
- Add Italian and English translations.
- Keep the version archiving dialog unchanged.
- Add focused test coverage for the new guidance.

## 🧪 How to test

- Open an active e-service from the producer's “My e-services” area and select “Archive e-service”.
- Verify that the notice period section explains that, when a version is already being archived, the entire e-service can only be archived after that version's scheduled archiving date.
- Open the archive dialog for a single version and verify that this additional guidance is not displayed.

[PIN-10659]: https://pagopa.atlassian.net/browse/PIN-10659?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/dialogs/DialogArchiveEservice.tsx
+++ b/src/components/dialogs/DialogArchiveEservice.tsx
@@ -79,7 +79,7 @@ const DialogArchiveEservice: React.FC<DialogArchiveEserviceProps> = ({ eserviceI
               <Typography variant="body2" sx={{ whiteSpace: 'pre-line' }}>
                 {t('content.advice.description')}
               </Typography>
-              <GracePeriodField />
+              <GracePeriodField description={t('content.advice.gracePeriodDescription')} />
             </Stack>
           )}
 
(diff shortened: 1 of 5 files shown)


antofo (commented):
Is it right that the `gracePeriodDescription` text is shown even if the e-service has only one version?

DenisLantero:
> Is it right that the `gracePeriodDescription` text is shown even if the e-service has only one version?

In my opinion it makes sense to show the description even if the eservice has only one version, I'll check with the design team and let you know

borgesis95:
Good catch! @antofo just a context about this: It is a temporary message will be replaced soon with an error message in next release.

antofo (approved):

borgesis95 (approved):

PR #2147, opened 2026-07-28
fix: fix function for last version and add tests (PIN-10548)

alten-dturus:
## 🔗 Issue

[PIN-10548](https://pagopa.atlassian.net/browse/PIN-10548)

## 📝 Description / Context

This PR fix the function so a provider can go to the last version even if the versions are archived.

## 🛠 List of changes

- add `isProviderOrConsumer` prop in the function `getViewLatestVersionTargetId` so the version states are different from provider and consumer
- add tests

[PIN-10548]: https://pagopa.atlassian.net/browse/PIN-10548?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/utils/eservice.utils.ts
+++ b/src/utils/eservice.utils.ts
@@ -61,10 +61,10 @@ export function getActiveDescriptor(descriptors: Array<CompactDescriptor> | unde
 export function getViewLatestVersionTargetId(
   descriptors: Array<CompactDescriptor> | undefined,
   currentDescriptorId: string | undefined,
-  isProviderOrConsumer: 'provider' | 'consumer'
+  providerOrConsumer: 'provider' | 'consumer'
 ) {
   const excludedStates: Array<EServiceDescriptorState> =
-    isProviderOrConsumer === 'provider'
+    providerOrConsumer === 'provider'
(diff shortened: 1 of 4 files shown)


martinaCampoli (approved):

src/utils/eservice.utils.ts
```diff
@@ -60,12 +60,16 @@ export function getActiveDescriptor(descriptors: Array<CompactDescriptor> | unde
 
 export function getViewLatestVersionTargetId(
   descriptors: Array<CompactDescriptor> | undefined,
-  currentDescriptorId: string | undefined
+  currentDescriptorId: string | undefined,
+  isProviderOrConsumer: 'provider' | 'consumer'
```
antofo, line 64: This works properly and it is not a big deal, but I will suggest to change the name of this argument to something like `organizationType`, the `isProviderOrConsumer` suggest that the variable may be a boolean flag.

antofo (approved):

PR #2202, opened 2026-09-03
feature/10872: cant delegate an e-service that is in ARCHIVED state

emabruz94:
## 🔗 Issue

[PIN-10872](https://pagopa.atlassian.net/browse/PIN-10872)

## 📝 Description / Context

Can't delegate an e-service that is in ARCHIVED state, so we don't need to show it in the autocomplete

## 🛠 List of changes

- added `states` queryParams to getProviderList not including the `ARCHIVED` state

## 🧪 How to test

Go to: Il tuo ente > Gestione delle deleghe > Crea delega

[PIN-10872]: https://pagopa.atlassian.net/browse/PIN-10872?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/pages/DelegationCreatePage/components/__tests__/DelegationCreateEServiceAutocomplete.test.tsx
+++ b/src/pages/DelegationCreatePage/components/__tests__/DelegationCreateEServiceAutocomplete.test.tsx
@@ -70,7 +70,7 @@ describe('DelegationCreateEServiceAutocomplete', () => {
     >)
   })
 
-  it('uses the full set of provider states for a producer delegation', () => {
+  it('filters out ARCHIVED e-services from the provider autocomplete for a producer delegation', () => {
     renderWithApplicationContext(
       <DelegationCreateEServiceAutocomplete delegationKind="DELEGATED_PRODUCER" />,
       { withReactQueryContext: true }
--- a/src/pages/DelegationCreatePage/components/__tests__/DelegationCreateEServiceAutocomplete.test.tsx
+++ b/src/pages/DelegationCreatePage/components/__tests__/DelegationCreateEServiceAutocomplete.test.tsx
@@ -1,0 +1,108 @@
+import React from 'react'
+import { useQuery } from '@tanstack/react-query'
+import { vi } from 'vitest'
+import { renderWithApplicationContext } from '@/utils/testing.utils'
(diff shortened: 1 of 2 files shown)


src/pages/DelegationCreatePage/components/__tests__/DelegationCreateEServiceAutocomplete.test.tsx
```diff
@@ -0,0 +1,108 @@
+import React from 'react'
```
borgesis95, line 1: Without thinking about coverage: In this case do you believe that this test could be useful for our scope? We would like to test if when you open Autocomplete you don't want see `ARCHIVED` e-service. This seems not to be our responsability (?) . If we want do add this test an idea could be test if within state ARCHIVED is not contained 

```
const { mockedGetProviderList, mockedGetCatalogList } = vi.hoisted(() => ({
  mockedGetProviderList: vi.fn((_params?: { states?: string[] }) => ({
    queryKey: ['provider-list'],
    queryFn: vi.fn(),
  })),
  mockedGetCatalogList: vi.fn(() => ({ queryKey: ['catalog-list'], queryFn: vi.fn() })),
}))
```

test looks like:

```
 it('Should requests the provider e-service list excluding the ARCHIVED state for a producer delegation', () => 
{
    renderWithApplicationContext(
      <DelegationCreateEServiceAutocomplete delegationKind="DELEGATED_PRODUCER" />,
      { withReactQueryContext: true }
    )

    const states = mockedGetProviderList.mock.calls[0]?.[0]?.states
    expect(states).toBeDefined()
    expect(states).not.toContain('ARCHIVED')
  })
)
```

emabruz94: This test isn't actually an integration test. We don't wanna check if the results contain the ARCHIVED e-services cuz that isn't our responsibility indeed, its purpose and our responsibility is only to test and ensure that those two queries are correctly set up and those api get called with the correct queryParams

antofo (approved):

borgesis95 (approved):

l-giacomini (approved):

